PR #8, opened 2025-01-11
refactor: reorganize for better workflows

sroussey:
[x] Move examples away from other packages
[x] Move providers out of core in prep for webworkers
[x] Move cache providers out of core (except memory)
[x] Move AI/LLM/etc specific tasks out of core


sroussey:
Closes [#9](https://github.com/sroussey/ellmers/issues/9)

PR #17, opened 2025-01-16
Refactor KV Storage and Setup Models data to be stored

sroussey:
Very much work in process. Stuff is broken. But time to have the list of models not (always) hard coded.


sroussey:
Failed tests are a bug in current version of bun

PR #68, opened 2025-03-02
type.constructor(input, config)

sroussey:


sroussey (commented):
Very early, typescript, and particularly arrays are very unhappy. tests run fine though. not so sure about examples

PR #407, opened 2026-04-11
Add privateResourceScopes to enforce SSRF scope on redirects

sroussey:
## Summary

This PR adds a new `privateResourceScopes` option to `SafeFetchOptions` that enforces the task's declared `network:private` entitlement scope on every redirect hop, preventing SSRF attacks where a compromised upstream server uses Location headers to pivot to unauthorized private hosts or ports.

## Key Changes

- **New `urlMatchesScope()` function** in `UrlClassifier.ts`: Canonicalizes URLs and checks if they match any pattern in a list of resource patterns using glob matching. Returns false on parse errors (fail-closed).

- **Extended `SafeFetchOptions` interface**: Added optional `privateResourceScopes` field that accepts an array of glob patterns. When defined alongside `allowPrivate: true`, it restricts private-host requests (including redirects) to only those patterns.

- **Scope enforcement in `defaultSafeFetch()`** (browser path): Updated `assertAllowedUrl()` to check redirect targets against `privateResourceScopes` before following them, throwing `PermanentJobError` if out of scope.

- **Scope enforcement in `serverSafeFetch()`** (Node/Bun path): Added identical scope validation in `fetchOneHop()` to enforce the same restrictions on the server-side redirect loop.

- **FetchUrlJob integration**: Modified `FetchUrlJob.execute()` to thread `privateResourceScopes` derived from the input URL's `urlResourcePattern()` into the `fetchWithProgress()` call. This ensures the runtime scope matches what the task declared in its `entitlements()` method.

- **Comprehensive test coverage**: Added 50+ new tests covering:
  - `urlMatchesScope()` unit tests (edge cases: IPv6, hostname-prefix attacks, case normalization, invalid URLs)
  - Scope enforcement via stub (plumbing verification)
  - Real redirect loop enforcement (7 scenarios: in-scope redirects, cross-host/port/scheme pivots, chained redirects, case normalization, legacy behavior)
  - FetchUrlJob integration (scope threading from declared entitlements)

## Implementation Details

- **Backward compatible**: When `privateResourceScopes` is `undefined`, the boolean `allowPrivate` governs alone, preserving legacy behavior for direct callers outside the task layer.
- **Fail-closed**: URL parse errors in `urlMatchesScope()` return false, preventing accidental allowlisting of malformed URLs.
- **Canonicalization**: All URLs are normalized via `new URL(url).toString()` before matching, so minor syntactic differences (uppercase host, default ports) cannot bypass checks.
- **Consistent enforcement**: Both browser and server paths apply the same scope check on every redirect hop, ensuring the task cannot be tricked into reaching unauthorized private origins.

https://claude.ai/code/session_0134L6Tzad2N6k4nXeWGPbY7

### DIFF
--- a/packages/tasks/src/task/FetchUrlTask.ts
+++ b/packages/tasks/src/task/FetchUrlTask.ts
@@ -126,7 +126,10 @@ export type FetchUrlTaskOutput = FromSchema<typeof outputSchema>;
 
 async function fetchWithProgress(
   url: string,
-  options: RequestInit & { allowPrivate?: boolean } = {},
+  options: RequestInit & {
+    allowPrivate?: boolean;
+    privateResourceScopes?: readonly string[];
+  } = {},
   onProgress?: (progress: number) => Promise<void>
 ): Promise<Response> {
   if (!options.signal) {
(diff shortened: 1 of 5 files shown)


sroussey:
@copilot apply changes based on the comments in [this thread](https://github.com/workglow-dev/workglow/pull/407#pullrequestreview-4094120292)

PR #426, opened 2026-04-23
Optimize same-process job queue with direct notifications and drain-on-stop

sroussey:
## Summary
This PR optimizes the job queue system for same-process deployments (where a `JobQueueClient` is attached to a `JobQueueServer`) by introducing direct in-memory communication paths and improving graceful shutdown behavior. It eliminates unnecessary storage subscriptions and polling delays for local job submissions and aborts.

## Key Changes

- **Direct job notification**: When a `JobQueueClient` submits a job, it now calls `server.handleJobAdded()` directly instead of relying on storage subscriptions or poll intervals. This allows workers to pick up jobs immediately without waiting for the poll interval.

- **In-process abort handling**: Added `requestAbort()` method to `JobQueueWorker` that fires abort controllers directly. The client's `abort()` method now tries the local path first via `server.abortJob()` before falling back to storage writes for cross-process scenarios.

- **Graceful drain-on-stop**: Implemented a two-phase shutdown in `JobQueueWorker.stop()`:
  - Phase 1: Wait up to `stopTimeoutMs` (default 30s) for in-flight jobs to complete naturally
  - Phase 2: Force-abort any remaining jobs and wait briefly for cleanup
  - Tracks in-flight jobs using a `Map<jobId, Promise>` that resolves when each job settles

- **Deferred job optimization**: Added `hasDeferredJobs` flag to skip redundant storage peeks in `getIdleDelay()` when there are no jobs scheduled for the future. The flag is set by `rescheduleJob()` when a deferred job is observed and cleared once a peek confirms nothing is pending.

- **Progress update optimization**: Removed storage write from `updateProgress()` — mid-job progress is now delivered only via in-memory events. Storage is only touched at terminal transitions (complete/fail/retry/abort), reducing write overhead for same-process deployments.

- **Conditional storage subscription**: The server now skips subscribing to storage changes when at least one client is attached, since those clients will call `handleJobAdded()` directly. This avoids unnecessary subscription overhead on backends like SQLite/Postgres.

- **Conditional cleanup loop**: The cleanup loop only starts if at least one retention TTL is configured, avoiding unnecessary periodic work when no cleanup is needed.

- **New configuration option**: Added `stopTimeoutMs` to `JobQueueWorkerOptions`, `JobQueueServerOptions`, and `JobQueueFactoryOptions` to control drain timeout behavior.

## Implementation Details

- In-flight job tracking uses `Promise.withResolvers()` to create promises that resolve (never reject) when jobs settle, allowing `Promise.allSettled()` to wait for all jobs without error handling complexity.
- The abort path avoids race conditions by skipping the `storage.abort()` write when the abort fires locally, preventing last-writer-wins conflicts on async storages.
- Deferred job detection is conservative: once a deferred job is observed, the flag stays true until a peek confirms nothing is pending in the future.

https://claude.ai/code/session_01UJr5y7wV4ehvFQhaYpZhWh


sroussey:
@copilot rebase on main

sroussey:
@copilot+gpt-5.4 rebase on main again

sroussey:
@copilot rebase on main

sroussey:
@copilot resolve the merge conflicts in this pull request

sroussey:
@copilot rebase on main

PR #500, opened 2026-05-14
fix(util/worker, ai/task): TTL-based pendingAborts eviction; clarify runWithIterable bond

sroussey:
## Why

`WorkerServerBase.pendingAborts` previously used FIFO eviction at size 1000 (dropping the oldest 500 in bulk once the cap was crossed). A misbehaving client spamming aborts for ids that never landed could silently evict a legitimate pending abort before its matching `call` arrived &mdash; the run-fn would then execute un-aborted, defeating the whole abort-before-call race fix.

Separately, `runWithIterable`'s top JSDoc misleadingly suggested a two-way abort bond ("leaves the parent alone"). The implementation is actually parent &rarr; child only, and the wording invited future contributors to reintroduce reverse propagation by accident.

## What

**TTL-based pendingAborts** (`packages/util/src/worker/WorkerServerBase.ts`):
- `pendingAborts` is now `Map<string, number>` keyed on insertion timestamp instead of `Set<string>`.
- 30s TTL via `PENDING_ABORT_TTL_MS`. Each insertion schedules a one-shot `setTimeout` cleanup (mirrors the existing `scheduleCompletedRequestCleanup` pattern); timer handles are tracked in `pendingAbortTimers` so dispose-style hooks can clear them.
- Inline TTL sweep on every `recordPendingAbort` &mdash; drops any entries older than the TTL.
- 10k hard cap (`PENDING_ABORT_HARD_CAP`) as a memory safety-net; when exceeded, evict the half with the lowest timestamps (oldest-by-timestamp, not insertion order).
- `consumePendingAbort` rechecks the TTL on consumption: present + fresh &rarr; abort + delete + return `true`; present + expired &rarr; delete + return `false`; absent &rarr; return `false`.

**Comment fix** (`packages/ai/src/task/base/runWithIterable.ts`):
- Replace the misleading paragraph with explicit one-way-bond description plus two `@example` blocks showing parent-cancel vs consumer-break semantics.
- Add an inline `// One-way bond: parent -> child only.` marker above the `addEventListener("abort", ...)` line.

**Tests** (`packages/test/src/test/util/WorkerServerBase.race.test.ts`):
- New `describe("WorkerServerBase pendingAborts eviction policy", ...)` block with three cases under `vi.useFakeTimers()`:
  - **Eviction-correctness**: 1500 distinct aborts in a single tick + matching `call` for `a-500` &rarr; signal observed aborted (would have failed under FIFO).
  - **TTL-expiry**: single abort + `vi.advanceTimersByTime(31_000)` + matching `call` &rarr; signal observed un-aborted.
  - **Hard-cap safety**: `HARD_CAP + 10` aborts with 1ms steps &rarr; `pendingAborts.size <= HARD_CAP`, newest id survives (consume succeeds), oldest id evicted (consume fails). Also asserts the behavioural path via the same run-fn registration shape as the existing tests.

## Verification

- `bun test packages/test/src/test/util/WorkerServerBase.race.test.ts` &mdash; runs the three new tests plus the three pre-existing abort-before-call race tests (regression coverage).
- Type-check + lint on the two source files.

## Risks

- The 30s TTL must comfortably exceed worst-case message-port roundtrip latency between the main thread and the worker. In practice this is microseconds-to-milliseconds, so 30s is a 10,000x safety margin &mdash; not a concern.
- `consumePendingAbort` now rechecks `Date.now() - ts > TTL` on every call (one extra subtraction). Negligible.
- The per-id `setTimeout` adds one timer per abort. With the 10k hard cap that's at most 10k pending timers &mdash; well within Node/Bun limits, and they GC as they fire.

## Rollback

Two independent commits, each rollback-able on its own:
1. `fix(util/worker): TTL-based pendingAborts eviction` &mdash; revert restores the FIFO Set policy and removes the new tests.
2. `docs(ai/task): clarify runWithIterable abort bond is one-way` &mdash; comment-only; revert restores the prior JSDoc verbatim.

---

Note: this PR is branched off `main` as `claude/loving-mendel-Hcznj-abort` rather than the default session branch `claude/loving-mendel-Hcznj` because a separate PR re-enabling CI is being authored on that branch in parallel. The two changes have independent review concerns and should be reviewable in isolation.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LkKCHC8q6b5e7Nwvc8L3M6)_


sroussey:
@copilot resolve the merge conflicts in this pull request

PR #514, opened 2026-05-20
Chrome-ai

sroussey:
- Added support for `@types/dom-chromium-ai` to improve type definitions for Chrome AI APIs.
- Introduced a new `WebBrowser_Chat` function to handle multi-turn chat sessions, allowing for better session management and context retention.
- Implemented structured generation capabilities with `WebBrowser_StructuredGeneration`, enabling JSON output from the AI model.
- Enhanced tool-calling functionality with `WebBrowser_ToolCalling`, allowing the model to invoke tools and handle their results seamlessly.
- Updated existing capabilities to include new features such as `json-mode` and `tool-use`, expanding the range of tasks the AI can perform.
- Refactored session management with `WebBrowser_Sessions` to cache and manage AI sessions effectively, improving performance and resource utilization.
- Removed deprecated type definitions from `WebBrowser_ChromeAI.d.ts` to streamline the codebase.

This commit significantly enhances the functionality and usability of the Chrome AI provider, paving the way for more complex interactions and improved user experience.


sroussey:
## HIGH-priority review findings — fix plans

Following an end-of-day automated review of PRs updated in the last 24h. 5 HIGH findings, 0 CRITICAL. Recommend addressing H1–H4 before merge; H5 is a dead-code cleanup that can be deferred. Full detail per finding below.

### H1 — `WebBrowser_StructuredGeneration` defeats `StructuredGenerationTask.maxRetries`

**Where:** `providers/chrome-ai/src/ai/common/WebBrowser_StructuredGeneration.ts:175-195`

**Why:** The run-fn throws `PermanentJobError` on JSON.parse or schema-validation failure. The inline comment claims `StructuredGenerationTask` runs inside a retry loop that catches per-attempt errors — but `packages/ai/src/task/StructuredGenerationTask.ts:177` consumes the inner stream via `for await (const event of super.executeStream(...))`, which lets exceptions propagate straight out of the generator. The task's retry/repair pipeline only runs when the run-fn emits a `finish` event whose `data.object` parses but fails validation. Chrome's path therefore silently disables `maxRetries`.

**Fix:** Mirror `Anthropic_StructuredGeneration_Stream` / `HFT_StructuredGeneration_Stream`. On parse failure, fall back to `parsePartialJson(accumulatedJson) ?? {}`. On validation failure, do NOT throw — still emit `{ type: "finish", data: { object: finalObject } }` and let the task's compiled validator decide retry. Compile-time invalid-schema (the `PermanentJobError` at lines 113-118) legitimately stays as throw.

**Effort:** S (~25 lines).

### H2 — `WebBrowser_ToolCalling` cache reuse double-feeds tool results

**Where:** `providers/chrome-ai/src/ai/common/WebBrowser_ToolCalling.ts:200-260` (cache check), `:102-118` (trade-off comment)

**Why:** The run-fn caches `LanguageModel` and reuses across orchestrator turns when `input.messages` is present, gated on `toolsFingerprint` and `priorMessageCount`. Chrome's `promptStreaming()` internally appends tool-result turns to the session's history. On the orchestrator's next turn, `input.messages` re-supplies the tool result as a text frame, while the cached session also has its own appended copy → model sees the same tool result twice. The trade-off comment acknowledges the failure mode; the chosen `priorMessageCount` watermark does not actually defend against it.

**Fix:** Delete the cache-reuse path for tool-calling entirely. Match `WebBrowser_TextGeneration.ts`'s pattern (always `factory.create` + `session.destroy` in `finally`). Remove `cached` / `usedCachedSession` / `cacheable` / `cacheWritten` locals and the `getChromeSession` / `setChromeSession` / `dropChromeSessionEntry` imports. Chrome session-create is cheap on a warm model; the correctness win outweighs the amortization. If amortization matters later, gate on `capturedCalls.length === 0` from the previous turn (only safe when the prior turn was text-only).

**Effort:** S (~40 lines deletion).

### H3 — `WebBrowser_Chat` history-replay watermark drifts from Chrome's effective initial prompt

**Where:** `providers/chrome-ai/src/ai/common/WebBrowser_Chat.ts:67-78`, `providers/chrome-ai/src/ai/common/WebBrowser_ChatHistory.ts:38-67`

**Why:** Cache watermark is `priorHistory.length` (raw count of `ChatMessage`s). But `buildInitialPromptsFromHistory` silently filters out empty-text messages, tool-role messages, and mid-history system messages. Two histories that produce identical `initialPrompts` but differ in dropped-frame count force a needless rebuild (cache miss). Conversely, a mid-conversation mutation of the leading system text — also silently swallowed by the filter — keeps the cache hot with the old system prompt baked in.

**Fix:** Compute the watermark over the FILTERED history that actually got fed to Chrome. Change `buildInitialPromptsFromHistory` to return `{ initialPrompts, fingerprint }` where `fingerprint` is `canonicalStringify(initialPrompts)` (or a hash). Gate cache reuse on the fingerprint match, not raw length. Add a separate leading-system fingerprint check so mid-conversation system-prompt mutations force a rebuild. Update `ChromeChatSessionState` (`WebBrowser_Sessions.ts:17-34`) to carry `historyFingerprint?: string` alongside the existing `messageCount` (preserved for tool-calling/structured-gen which don't change in this fix).

**Effort:** M (~80 lines + interface touch).

### H4 — Bare `LanguageModel` (and friends) references can throw at module load on non-Chrome

**Where:** Eight files — `WebBrowser_Chat.ts:40-44`, `WebBrowser_StructuredGeneration.ts:100-104`, `WebBrowser_ToolCalling.ts:159-163`, `WebBrowser_TextGeneration.ts:42-46`, `WebBrowser_TextLanguageDetection.ts`, `WebBrowser_TextRewriter.ts`, `WebBrowser_TextSummary.ts`, `WebBrowser_TextTranslation.ts`

**Why:** Each file uses `typeof LanguageModel !== "undefined" ? LanguageModel : undefined`. The `typeof X` form is correct in classic scripts, but in strict-mode ES modules under some bundlers the second reference to `LanguageModel` (after the `?`) is treated as a binding lookup and can throw `ReferenceError` at module evaluation on Firefox/Safari/Node. `WebBrowser_CapabilityProbe.ts:80-90` already routes through `globalThis` and got this right — the probe fix just didn't propagate to the run-fns.

**Fix:** Add a tiny helper in `WebBrowser_ChromeHelpers.ts`:
```ts
export function getChromeGlobal<T = unknown>(name: string): T | undefined {
  return (globalThis as unknown as Record<string, T | undefined>)[name];
}
```

Replace each call site:
```ts
// before
getApi("LanguageModel", typeof LanguageModel !== "undefined" ? LanguageModel : undefined)
// after
getApi("LanguageModel", getChromeGlobal<typeof LanguageModel>("LanguageModel"))
```

Apply to LanguageModel, Summarizer, Rewriter, Translator, LanguageDetector across all 8 files.

**Effort:** S (~10 lines per site × 8 + helper).

### H5 — `WebBrowser_StructuredGeneration` cache code is dead in production

**Where:** `providers/chrome-ai/src/ai/common/WebBrowser_StructuredGeneration.ts:125-160` (cache branch); `packages/ai/src/task/StructuredGenerationTask.ts` (no sessionId plumbing)

**Why:** The fingerprint and cache infrastructure adds ~80 lines plus three live failure modes (cache poisoning, schema-mismatch invalidation, watermark drift). But `StructuredGenerationTask` doesn't allocate a `sessionId` and the JobInput plumbing that produces one is owned by `AiChatTask`. No production caller passes `sessionId`. Dead in production.

**Fix:** Delete the cache code in this file (always create + destroy like `WebBrowser_TextGeneration`). Leave a TODO comment pointing at the follow-up: wiring `StructuredGenerationTask` to allocate a sessionId for json-mode providers is a cross-provider contract change and belongs in a separate PR. If `canonicalStringify` is needed by the H3 fix, promote it to a shared helper; otherwise inline-delete.

**Effort:** S (~90 lines deletion).

### Sequencing

- H1, H2, H4 are independent — any order.
- H1 + H5 both touch `WebBrowser_StructuredGeneration.ts` — land H5 first (deletes more), then H1 on the simplified file.
- H3 touches a shared `ChromeChatSessionState` interface that H2 reads (`messageCount`); H3's plan keeps `messageCount` and adds `historyFingerprint` as an optional sibling, so the two don't conflict.
- If H4 lands first, H1/H5 pick up `getChromeGlobal` for free.

---

sroussey:
@copilot+gpt-5.4 Some of this was recently added to main branch, but not all. It needs to be rebased and checked carefully. main has some work on sessions and making sure that streaming does not use snapshot.

sroussey:
@claude Some of this was recently added to main branch, but not all. It needs to be rebased and checked carefully. main has some work on chrome sessions too and making sure that chrome ai streaming does not send snapshot when they aren't actually snapshots. Double check docs at google. Text generation and translation may be different.

PR #557, opened 2026-06-09
fix(task-graph): harden CacheRef, canonicalize binary format, byte-bounded backpressure

sroussey:
Three follow-up fixes for the binary-streaming framework + result-as-reference work in #545. Targets `claude/stoic-bell-RLJWT` so the patches merge with the rest of that PR.

## Fix 1 (CRITICAL) — Brand `CacheRef` with a literal `kind`

Files: `packages/task-graph/src/cache/CacheRef.ts`, `packages/task-graph/src/task/CacheCoordinator.ts`, `packages/task-graph/src/cache/RunPrivateCacheRepo.ts` + test updates.

`CacheRef` was discriminated by shape alone: any object with `{$ref: string}` satisfied `isCacheRef`. The cache-ref resolver walks task outputs and calls `getOutputByRef(ref)` on every match — so any code path surfacing an attacker-influenced `{$ref: "cache://OTHER_RUN/secret"}` (tool result, parsed JSON, AI structured output, embedded JSON-Schema reference) could trick `resolveJobOutput` / `resolveOutput` into reading bytes from another run or tenant's private cache slot.

Adds a literal `kind: "task-graph/CacheRef"` brand (a string, not a Symbol — so it survives JSON round-trip across queue rows / IPC), checked alongside `$ref` in `isCacheRef`. New `makeCacheRef(...)` helper constructs branded refs; `CacheCoordinator.getBinaryRefSinksByPolicy` and `RunPrivateCacheRepo.saveOutputStream` defensively re-wrap legacy backings.

## Fix 2 (HIGH) — Canonical `BinaryFormat` vocabulary

Files: `packages/task-graph/src/task/StreamTypes.ts`, `StreamProcessor.ts`, `CacheCoordinator.ts`, `TaskRegistry.ts` + test updates.

`materializeBinary` previously accepted any string and silently coerced unknown values (including casing typos like `"Blob"`) to the ArrayBuffer branch — a `format: "Blob"` mistake produced ArrayBuffers where every downstream consumer expected Blobs, with the mismatch only surfacing far from the task definition.

Introduces `BinaryFormat = "blob" | "binary"` and a single `assertBinaryFormat(schema, port)` helper used by `StreamProcessor`, `CacheCoordinator.hydrateRefsBelowThreshold`, and `TaskRegistry.registerTask`. Invalid `format` on an `x-stream: "binary"` port now throws at registration time, not during a streaming run.

## Fix 3 (HIGH) — Byte-bounded backpressure (default 8 MiB)

[...]

PR #558, opened 2026-06-10
fix(ai): tighten content-block types + add nightly schema-vs-type drift guard

sroussey:
## Summary

Follow-up hardening for PR #555. Tightens three runtime input types in `@workglow/ai` to eliminate (or document) drift between hand-written types and the JSON schemas they shadow, and adds a nightly type-drift guard so future schema or `ContentBlock` changes can't drift silently.

This PR targets `main` and is independent of PR #557 (which targets a different base).

## Per-finding rationale

**H-1 — `AiChatTaskInput['prompt']` and `ToolCallingTaskInput['prompt']`**

- `AiChatTask.ts` previously inlined the four `ContentBlock` variants as a literal union. Replaced with `string | readonly ContentBlock[]` so the runtime type reuses `ContentBlock` directly and a change to `ContentBlock` flows through automatically. The schema's items are already `ContentBlockSchema`, so this is the canonical match.
- `ToolCallingTask.ts` previously used `string | (string | { type: "text" | "image" | "audio"; [x]: unknown })[]` — a looser shape mirroring the JSON schema. Replaced with `string | readonly (string | ContentBlock)[]`. This is **intentionally tighter than the schema** (which still uses `additionalProperties: true` for UI flexibility); the drift test pins this divergence so a future schema broadening or `ContentBlock` change cannot drift silently.

**H-2 — `ChunkRetrievalTaskInput`**

The schema's `if/then/else` (when `query: string`, `model` is required) is invisible to `json-schema-to-ts` / `FromSchema`, so the hand-written discriminated union is intentionally stricter than what `FromSchema` would resolve to. Renamed `inputSchema` to the exported `ChunkRetrievalInputSchema` (so the drift test can reference it) and added a JSDoc block above the type explaining the intentional divergence. No behavior change.

**H-3 — Drift guardrail**

- `packages/ai/src/task/__tests__/types.test-d.ts` — vitest typecheck-mode tests using built-in `expectTypeOf`. Three assertions:
  - **AiChat**: positive assertions that the prompt type accepts every `ContentBlock` variant (`text`, `image`, `tool_use`, `tool_result`) and the union itself, plus `FromSchema<typeof AiChatInputSchema>['prompt']` resolves to a non-trivial type. This is the right drift signal — `FromSchema`'s resolution differs from the runtime type in non-meaningful ways (mutable array vs `readonly`, `is_error?:` vs `is_error: boolean | undefined`), so structural equality is too brittle.
  - **ToolCalling**: `not.toEqualTypeOf` — pins the intentional divergence. If the schema or `ContentBlock` ever bring the types into equality, the author should re-verify whether the tightening is still desired.
  - **ChunkRetrieval**: `not.toEqualTypeOf` — pins the `if/then/else` divergence. The moment `FromSchema` honours `if/then/else`, this test fails and the runtime type can switch to a `FromSchema`-derived form.
- `.github/workflows/nightly-typecheck.yml` — nightly + `workflow_dispatch` job that runs the drift test via `bunx vitest run --typecheck --typecheck.only`. Kept separate from `test.yml` so per-PR signal stays focused; the nightly catches drift the moment it lands on `main`.

A late commit (`ci(ai): run drift guard in vitest typecheck mode`) corrects the workflow command — `.test-d.ts` files are picked up by vitest's typecheck include pattern (`*.{test,spec}-d.ts`), not its default runtime include, so `--typecheck --typecheck.only` is required.

## Test plan

- [x] `bunx vitest run --typecheck --typecheck.only packages/ai/src/task/__tests__/types.test-d.ts` — 3/3 tests pass
- [x] `bun scripts/test.ts ai vitest` — 36 files / 238 tests pass
- [x] `bun run typecheck:budget` — `OK (34 packages within budget)`; `packages/ai` instantiation count is flat vs baseline (the substitutions reuse already-resolved types, no extra instantiation)
- [ ] Nightly workflow lands on `main` after merge; the next scheduled run (or a manual `workflow_dispatch`) verifies CI-side

https://claude.ai/code/session_01562Z29a2UQDNBVAcJGyUoY

---
_Generated by [Claude Code](https://claude.ai/code/session_01562Z29a2UQDNBVAcJGyUoY)_

### DIFF
--- a/.github/workflows/nightly-typecheck.yml
+++ b/.github/workflows/nightly-typecheck.yml
@@ -21,4 +21,6 @@ jobs:
         with:
           bun-version: latest
       - run: bun i
-      - run: bunx vitest run packages/ai/src/task/__tests__/types.test-d.ts
+      # `.test-d.ts` files only run in vitest's typecheck mode. `--typecheck.only`
+      # disables the runtime test run so this stays cheap.
+      - run: bunx vitest run --typecheck --typecheck.only packages/ai/src/task/__tests__/types.test-d.ts
--- a/.github/workflows/nightly-typecheck.yml
+++ b/.github/workflows/nightly-typecheck.yml
@@ -1,0 +1,24 @@
+name: Nightly type-drift guard
+permissions:
+  contents: read
+on:
+  schedule:
+    - cron: "17 7 * * *"
+  workflow_dispatch:
+env:
+  CI: "true"
+  DO_NOT_TRACK: "1"
+  TURBO_TELEMETRY_DISABLED: "1"
+jobs:
+  type-drift:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 24
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+      - run: bun i
+      - run: bunx vitest run packages/ai/src/task/__tests__/types.test-d.ts
--- a/packages/ai/src/task/AiChatTask.ts
+++ b/packages/ai/src/task/AiChatTask.ts
@@ -159,23 +159,7 @@ export type AiChatTaskInput = Omit<
     maxIterations?: number | undefined;
     responseFormat?: "text" | "markdown" | undefined;
     model: string | ModelConfig;
-    prompt:
-      | string
-      | (
-          | { type: "text"; text: string }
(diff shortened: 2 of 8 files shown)


sroussey:
Closing in favour of PR #565. The README changes from this PR were already committed to `main` directly (`bc0918d`). The remaining applicable fixes (export `ChunkRetrievalInputSchema`, drift-guard test, tsconfig update, nightly CI workflow) are cherry-picked into #565 targeting `main`.

---

PR #559, opened 2026-06-10
fix(task-graph): legacy unbranded {$ref} read-side rewrap + binaryBackpressure default no-op

sroussey:
## Summary

Stacks two follow-ups on top of PR #557 so they merge together.

### H-1 — Read-side rewrap of legacy unbranded `{$ref}` cache rows

Pre-brand cache writes landed as `{ $ref, size?, mime? }` without the literal `kind` discriminator that #557 introduced. Subsequent `isCacheRef` probes then treat those rows as ordinary objects and skip threshold-based rehydration. We upgrade them in place — but only at schema-declared binary streaming ports — by reusing the same `getStreamingPorts(...).filter(p => p.mode === "binary")` set `hydrateRefsBelowThreshold` uses. Non-binary fields that legitimately carry a `{$ref: string}` shape (JSON-Schema refs, user metadata) are NEVER auto-promoted, so the shape-only collision risk #557 closes stays closed.

Why option (b) over a `__cv` bump: a `__cv` bump throws away every pre-brand cached binary write. Read-side rewrap upgrades those rows transparently.

The new `isLegacyUnbrandedCacheRefShape` helper is a strict "needs upgrade" predicate; it rejects already-branded refs and shapes whose `size`/`mime` hints have the wrong type. `CacheCoordinator.lookup` upgrades in place on cache hit, and `hydrateRefsBelowThreshold` accepts both branded refs and the pre-brand shape as defence in depth on the streaming finish-event path that bypasses lookup. `RunPrivateCacheRepo.getOutput` deliberately does NOT re-wrap — that would have to walk arbitrary nested fields without the schema in scope.

### H-2 — `binaryBackpressure` default no-op

`IExecuteContext.binaryBackpressure`'s JSDoc already promised a "no-op default when the runtime does not install a real backpressure source." Only `StreamProcessor` was installing the router-aware version, so non-streaming `execute()` paths handed tasks an `IExecuteContext` whose field was undefined — calling it threw `TypeError`. Install a module-private shared no-op in `TaskRunner.executeTask` and `WhileTaskRunner.executeTask` so the unconditional-call pattern actually costs nothing per call.

`executePreview`'s `IExecutePreviewContext` is `Pick<IExecuteContext, "own">` and intentionally has no `binaryBackpressure` slot — the preview path needs no change.

### Test coverage

- `CacheRef.legacyShape.test.ts` — Case A (threshold 0, ref upgraded); Case B (rehydratable via `getOutputByRef`); **Case C — the load-bearing scope guard:** a non-binary port whose cached value is `{$ref: "#/$defs/X"}` (a legitimate JSON-Schema ref) MUST NOT be touched, and `getOutputByRef` MUST NEVER be invoked. If Case C breaks, the fix is unsafe to ship.
- `TaskRunner.binaryBackpressure.test.ts` — `execute()` can `await ctx.binaryBackpressure()` without throwing; repeated calls stay cheap; `runPreview()` still works.
- `StreamingBackpressure.abortDrain.test.ts` — abort-while-parked through `StreamProcessor.run` settles within ~200ms emitting `stream_end` or `stream_error`; sibling routers' waiters wake independently; multiple parked pushes on the same router are all released by a single drain (exercises the `prev`/`res` linked-callback chain in `BinaryStreamRouter._awaitDrain` / `push`).

## Test plan

- [x] `bun scripts/test.ts graph vitest` — 839 / 839 tests pass across 86 files
- [x] `bunx tsc -b packages/task-graph` — clean
- [x] All three new test files pass standalone (15 / 15 tests)

This PR stacks on #557 and inherits its base path to main.

https://claude.ai/code/session_01562Z29a2UQDNBVAcJGyUoY

---
_Generated by [Claude Code](https://claude.ai/code/session_01562Z29a2UQDNBVAcJGyUoY)_

### DIFF
--- a/packages/task-graph/src/cache/CacheRef.ts
+++ b/packages/task-graph/src/cache/CacheRef.ts
@@ -52,6 +52,35 @@ export function isCacheRef(value: unknown): value is CacheRef {
 }
 
 /**
+ * Matches the pre-brand on-disk shape `{ $ref: string, size?, mime? }` without
+ * the `kind` discriminator. Used ONLY at declared binary streaming port slots
+ * by {@link CacheCoordinator} to upgrade legacy cache rows in place. Never
+ * call this on arbitrary nested fields — that would defeat the brand check
+ * and re-introduce the shape-only collision risk that the brand exists to
+ * close.
+ *
+ * An already-branded {@link CacheRef} is rejected (returns `false`) so callers
+ * can use this as a strict "needs upgrade" predicate without first filtering
+ * out the branded case.
+ */
+export function isLegacyUnbrandedCacheRefShape(
+  value: unknown
+): value is { $ref: string; size?: number; mime?: string } {
+  if (typeof value !== "object" || value === null) return false;
+  const candidate = value as {
+    readonly kind?: unknown;
+    readonly $ref?: unknown;
+    readonly size?: unknown;
+    readonly mime?: unknown;
+  };
+  if (candidate.kind === CACHE_REF_KIND) return false;
+  if (typeof candidate.$ref !== "string") return false;
+  if (candidate.size !== undefined && typeof candidate.size !== "number") return false;
+  if (candidate.mime !== undefined && typeof candidate.mime !== "string") return false;
+  return true;
+}
+
+/**
  * Construct a branded {@link CacheRef}. Cache backings MUST use this helper (or
  * spread `{kind: CACHE_REF_KIND, ...}` themselves) so the resulting ref carries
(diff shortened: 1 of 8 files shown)


sroussey:
Closing in favour of PR #565. The fixes in this PR (legacy unbranded `{$ref}` rewrap, `binaryBackpressure` no-op default) depend on the binary-streaming framework introduced in PR #557 / #545, which targets `claude/stoic-bell-RLJWT` and hasn't merged to `main` yet. These will flow in with that base branch.

---

PR #580, opened 2026-06-14
fix(storage,postgres,sqlite): close vector-shape validation gap in Postgres + Sqlite stores

sroussey:
## Summary

Closes the HIGH follow-up from the code review of #579. The previous fix only validated vector shape in `SupabaseVectorStorage`; `PostgresVectorStorage` and `SqliteVectorStorage` have the same write-path round-trip problem, and `PostgresVectorStorage.similaritySearch` had a generic `catch` that silently falls back to in-memory cosine — a wrong-dim or non-finite query there returned garbage instead of throwing.

- Lift `assertVectorShape` into the shared `@workglow/storage` package as a free function and re-export it from `common.ts`.
- Refactor `SupabaseVectorStorage` to use the shared helper (delete its private copy; behavior + error messages unchanged).
- Override `put` / `putBulk` on `PostgresVectorStorage` and `SqliteVectorStorage` so a malformed vector throws before the pgvector text encoder / JSON encoder ever sees it. `putBulk` validates the whole batch before delegating, so a single bad entry rejects the entire write.
- Insert `assertVectorShape(query, dim, "query")` as the first statement of `similaritySearch` on both backends — for Postgres this bypasses the silent in-memory fallback path in the `catch` block.
- Add PGlite-backed `PostgresVectorStorage.validation.test.ts` and `:memory:` better-sqlite3 `SqliteVectorStorage.validation.test.ts` covering short / long / NaN / Infinity / null / undefined vectors on `put`, `putBulk`, and `similaritySearch`, including a regression guard that `putBulk` does not partially write when one entry is malformed.

## Test plan

- [x] `bun run build:packages` — all packages build cleanly.
- [x] `bun run build:types --force` — all 36 type-check tasks pass.
- [x] `bunx vitest run packages/test/src/test/vector/PostgresVectorStorage.validation.test.ts packages/test/src/test/vector/SqliteVectorStorage.validation.test.ts` — 16/16 new tests pass.
- [x] `bunx vitest run packages/test/src/test/storage-vector/SupabaseVectorStorage.validation.test.ts` — pre-existing Supabase suite still passes after lifting the helper.
- [x] `bun scripts/test.ts storage vitest` — full storage suite (1400 tests across 51 files) passes; no regressions.
- [x] `bunx vitest run packages/test/src/test/vector` — full vector suite (48 tests) passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_019VmooqVjVM4iQjgN1AkuGu)_


sroussey:
Superseded by #589, which cherry-picks this commit (vector-shape validation for Postgres + Sqlite) onto current `main` as part of a cumulative fix integrating #580/#581/#583/#584/#587. Closing.

---

PR #581, opened 2026-06-15
fix(storage,indexeddb): extend vector validation to InMemory + IndexedDB; lift helper (HIGH)

sroussey:
## Summary

Closes the silent-corruption gap left by the original `assertVectorShape` rollout (#579 / #580): only Postgres, SQLite, and Supabase validated vector inputs. `InMemoryVectorStorage` and `IndexedDbVectorStorage` accepted malformed vectors (wrong dimension, NaN / Infinity entries, null / undefined) and stored them — test fixtures and browser deployments could quietly persist garbage that no `similaritySearch` could recover.

- Lifts a `validateVectorEntities(entities, vectorPropertyName, dimensions)` batch helper alongside `assertVectorShape` in `@workglow/storage`. Throws the same `StorageValidationError` on the first malformed entry so a `putBulk` rejects the whole batch before any backend call.
- Overrides `put` / `putBulk` on `InMemoryVectorStorage` and `IndexedDbVectorStorage` to validate up front. On IndexedDB the throw fires synchronously inside the awaited Promise so no IDB transaction ever opens — atomic rejection by construction.
- Prepends `assertVectorShape(query, vectorDimensions, "query")` to both `similaritySearch` paths so a bad query fails fast instead of returning a meaningless score.
- Refactors the Postgres / SQLite / Supabase `put` / `putBulk` overrides to use the new helper for parity (no behaviour change, just a single batch call).

## Verification

All commands run locally on this branch (Node 22 dev container, Bun 1.3.11):

| Command | Result |
| --- | --- |
| `bun scripts/test.ts storage vitest` | 53 files passed, 1416 tests passed, 13 skipped |
| `bun scripts/test.ts rag vitest` | 24 files passed, 225 tests passed |
| `npx vitest run packages/test/src/test/vector/` | 7 files passed, 64 tests passed (incl. 16 new) |
| `bun run build:types --force` | 36 tasks successful |
| `npx prettier --check` (changed files) | All files use Prettier code style |

## Test plan

- [x] New `InMemoryVectorStorage.validation.test.ts` covers short / long / NaN / Infinity / null / undefined on `put`, atomic batch rejection on `putBulk` (`getAll()` empty after throw), and wrong-dimension / NaN query rejection on `similaritySearch`.
- [x] New `IndexedDbVectorStorage.validation.test.ts` mirrors the same cases under `fake-indexeddb/auto`, asserting no row reaches the object store on `putBulk` rejection.
[...]

PR #583, opened 2026-06-16
fix(storage,indexeddb): make vector putBulk truly atomic — close PR #581 follow-up (HIGH)

sroussey:
## Summary

Closes the HIGH atomicity gap identified in the PR #581 code review.

PR #581 validated vector shape up front in `InMemoryVectorStorage.putBulk` / `IndexedDbVectorStorage.putBulk`, but then fell through to the inherited tabular `putBulk`, which runs writes via `Promise.all`. A non-shape failure mid-batch (PK collision, IDB `QuotaExceededError`, listener throw, structured-clone failure on a real browser) could leave rows 0..N-1 committed even though the PR description, CHANGELOG, and JSDoc promised an all-or-nothing batch.

This PR makes the guarantee real (Option B from the review thread), rather than softening the wording (Option A).

### `packages/storage`

- **`InMemoryTabularStorage`** — exposes `protected snapshotMutableState()` / `restoreMutableState()` so subclasses can roll back partial `Map` writes and the autoincrement counter without reaching into private state.
- **`InMemoryVectorStorage.putBulk`** — snapshots, writes serially via the inherited `put()`, and restores the snapshot on throw. The pre-existing rows survive, the failed batch's rows are absent, and the autoincrement counter is restored so the next successful `put` does not skip an id.
- **`TabularEventListeners`** — adds a `rollback` event carrying `{ op, error }` so subscribers can reconcile any per-row `put` events that landed before the failure. Existing event subscribers are unaffected.

### `packages/indexeddb`

- **`IndexedDbTabularStorage.prepareRecordForPut`** — extracted from `put` so the per-record key-derivation rules (autoincrement, UUID, `clientProvidedKeys`) live in one place and are shared by the single- and batch-write paths.
- **`IndexedDbTabularStorage.putBulkInTransaction`** (new `protected`) — routes every record through a single `readwrite` IDB transaction. Request errors, transaction aborts, and quota failures abort the whole batch; no row reaches the object store. Per-row `put` events are deferred to `tx.oncomplete` so subscribers see a burst commit rather than interleaved request notifications. An `InvalidStateError` raised by an `onversionchange` race is absorbed with a single open-and-retry, and a `rollback` event fires before the retry runs.
- **`IndexedDbVectorStorage.putBulk`** — now routes through `putBulkInTransaction`. The base class `putBulk` is unchanged so existing non-vector callers keep their per-row event semantics.

### Tests — `packages/test/src/test/vector/`

- **`InMemoryVectorStorage.atomicity.test.ts`** (4 tests)
  - `putBulk` rolls back all writes when a non-shape error fires mid-batch (spy on `InMemoryTabularStorage.prototype.put`, throw on the 2nd call, assert pre-existing row survives and batch rows are absent).
  - Emits a `rollback` event on the storage emitter when a batch fails.
  - `putBulk` commits every row when nothing throws.
  - `putBulk` does not advance the autoincrement counter on failure (proves restoration via `id === 2` on next successful put).
- **`IndexedDbVectorStorage.atomicity.test.ts`** (3 tests)
  - `putBulk` rolls back the entire batch when any IDB request fails (force `tx.abort()` on the 2nd request via an `IDBObjectStore.prototype.put` spy — `fake-indexeddb` does not enforce structured-clone, so the abort path is the right lever).
  - `putBulk` commits all rows on success and emits per-row `put` events after `tx.oncomplete` (in batch order).
  - Emits a `rollback` event on failure.

### CHANGELOGs

- `packages/storage/CHANGELOG.md` — `Unreleased`: real atomicity guarantee, snapshot/restore helpers, `rollback` event.
- `packages/indexeddb/CHANGELOG.md` — `Unreleased`: single-transaction batch, per-row `put` events deferred to `tx.oncomplete` (behaviour change for subscribers).

## Verification

Run from `/home/user/libs`:

| Command | Result |
| --- | --- |
| `bun scripts/test.ts storage vitest` | 51 files passed, 1391 tests passed, 13 skipped |
| `bun scripts/test.ts queue vitest` | 20 files passed, 488 tests passed, 26 skipped |
| `bun scripts/test.ts rag vitest` | 24 files passed, 225 tests passed |
| `bunx vitest run packages/test/src/test/vector/` | 5 files passed, 39 tests passed (incl. 7 new) |
| `bun run build:types --force` | 36 tasks successful |

## Test plan

- [x] New `InMemoryVectorStorage.atomicity.test.ts` covers mid-batch rollback, autoincrement counter restoration, `rollback` event emission, and the success path.
- [x] New `IndexedDbVectorStorage.atomicity.test.ts` covers single-transaction batch abort, `rollback` event emission, and `put` event ordering on success.
- [x] Existing storage suite (1391 tests) green — the base `IndexedDbTabularStorage.putBulk` is unchanged, only `IndexedDbVectorStorage.putBulk` is rerouted.
- [x] RAG suite green — RAG chunk ingestion goes through vector `putBulk` and the new transactional path does not regress any RAG fixture.
- [x] Job-queue suite green — confirms the IDB tabular `put` refactor (now using `prepareRecordForPut`) did not change semantics for IDB queue / rate-limiter stores.
- [x] `bun run build:types --force` clean across all 36 packages.

## Notes

This is an independent follow-up off `main`, NOT a cherry-pick of #581 nor a rebase onto its branch. The atomicity work is structurally orthogonal to #581's validation work: validation runs first to reject malformed inputs synchronously, and the new transactional / snapshot-restore code makes the post-validation write path actually all-or-nothing.

Option B was the right shape — `IndexedDbTabularStorage`'s base `putBulk` is preserved so non-vector callers keep their existing per-row event semantics, and the new `putBulkInTransaction` is a `protected` hook only the vector overlay opts into. No existing tests needed fixture changes.

https://claude.ai/code/session_01UPaYrAyt6Fibf9faFas6YR

---
_Generated by [Claude Code](https://claude.ai/code/session_01UPaYrAyt6Fibf9faFas6YR)_


sroussey:
Superseded by #589, which cherry-picks this commit (atomic vector `putBulk` — snapshot/restore + single-transaction IDB batch) onto current `main`, integrated so validation (#581) runs before the atomic write. Part of the cumulative fix for #580/#581/#583/#584/#587. Closing.

---

PR #584, opened 2026-06-17
fix(storage,indexeddb): safe-emit so a throwing listener does not mask batch errors or hang putBulk (HIGH follow-up to #583)

sroussey:
## Summary

Follow-up to #583. Code review of that PR surfaced a HIGH-severity issue: the four new storage event emit sites it added crash the caller when a listener throws, because `EventEmitter.emit` rethrows listener errors synchronously (or as an `AggregateError`).

In two of the four sites that signal turns into data loss; in one of them it deadlocks `putBulk`:

1. **`packages/storage/src/vector/InMemoryVectorStorage.ts` — `putBulk` rollback emit.** The `emit("rollback", ...)` runs *before* `throw error`. A throwing rollback listener masks the original batch error — callers think the batch succeeded.
2. **`packages/indexeddb/src/storage/IndexedDbTabularStorage.ts` — `tx.oncomplete` per-row `put` loop.** A throwing `put` listener aborts the loop *before* `resolve(results)`, so the `putBulk` promise hangs forever even though every row already committed to IDB. The caller never knows the writes landed.
3. **Same file, abort/error rollback emit.** Same masking hazard as (1).
4. **Same file, `InvalidStateError` retry-branch rollback emit.** Same masking hazard, plus this one can swallow the original error mid-retry.

## Fix shape

- New `packages/storage/src/events/safeEmit.ts` wraps `emitter.emit(...)` in a `try/catch`, logs listener errors via `getLogger().warn`, and re-throws on a microtask so observability (`unhandledException` / `unhandledrejection`) stays intact. The synchronous caller keeps running.
- The four PR #583 emit sites now route through `safeEmit(this.events, ...)`. The trailing `throw error` and `resolve(results)` calls now run unconditionally.
- `EventEmitter` is **unchanged** — the dozens of other emitters across the codebase keep their current rethrow semantics. `safeEmit` is opt-in for code paths where a misbehaving subscriber must not derail a write.
- `safeEmit` is re-exported from `@workglow/storage`'s public surface alongside the other event-related exports.

## Tests

Extended the two atomicity test files added by #583:

- **`InMemoryVectorStorage.atomicity.test.ts`**
  - `does not mask the original batch error when a rollback listener throws` — installs a throwing rollback listener + forces the underlying `put` to throw `"original-error"`, asserts `putBulk` still rejects with `"original-error"`.
  - `still fires later rollback listeners when an earlier one throws` — two rollback listeners; first throws, second records the payload; asserts the second still received `{ op: "putBulk", error }`.
- **`IndexedDbVectorStorage.atomicity.test.ts`**
  - `does not hang putBulk when a put listener throws on the success path` — registers a throwing `put` listener, then races `storage.putBulk([3 rows])` against a 2-second timeout. `putBulk` wins, returns 3 results, and the rows are present via `storage.get(...)`.
  - `still fires later put listeners when an earlier one throws on the success path` — two `put` listeners; first throws, second collects ids; asserts the second got all three ids in batch order.

Both test files also install `process.on("uncaughtException", () => {})` / `unhandledRejection` no-ops at file top so vitest doesn't fail on the `safeEmit` microtask rethrow (which is intentional — it's how observability tooling still sees the listener bug).

## Verification

Run from `/home/user/libs`:

- `bun scripts/test.ts storage vitest` — 51/51 files passed; 1395 tests passed, 13 skipped.
- `bun scripts/test.ts rag vitest` — 24/24 files passed; 225 tests passed.
- `bun run build:types --force` — 36/36 turbo tasks successful.
- `bun run format` — clean.

## Test plan

- [x] New tests assert the four behavioral guarantees above.
- [x] Existing `InMemoryVectorStorage` and `IndexedDbVectorStorage` atomicity tests still pass.
- [x] `EventEmitter` semantics for non-storage emitters are untouched (no `EventEmitter.ts` edits).
- [x] `build:types` is clean across all 36 packages.

Base intentionally set to `claude/beautiful-mayer-l9vwzg` (the parent PR's head) so reviewers see only this delta.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PaACxghT8otw2MvXopfPTE)_


sroussey:
Superseded by #589, which cherry-picks this commit (`safeEmit` so a throwing listener can't mask a batch error or hang `putBulk`) onto current `main`. Part of the cumulative fix for #580/#581/#583/#584/#587. Closing.

---

PR #587, opened 2026-06-18
fix(storage,util): vector rollback mutex + safeEmit channel + fail-closed credential resolver (4 HIGH from review)

sroussey:
## Summary

Addresses four HIGH-severity findings from code review of #583 / #584:

### Storage / vector

1. **`packages/storage/src/vector/InMemoryVectorStorage.ts` — mutex.** `putBulk` snapshots + restores on a mid-batch throw, but a concurrent single-row `put` could land between the snapshot and the rollback, then be wiped by the restore even though it had committed cleanly. The vector overlay's all-or-nothing guarantee only covered the batch's own rows. Adds a per-instance promise-chain mutex (mirroring the PGlite-path mutex in `PostgresTabularStorage`) so `put` and `putBulk` serialize on the same chain.

2. **`packages/storage/src/events/safeEmit.ts` — rethrow channel.** `safeEmit` previously forwarded a caught listener throw via `queueMicrotask(() => { throw e })`, which surfaces to Node as an `uncaughtException` — a signal that, if no handler is installed, defaults to terminating the process. Storage event subscribers should not be able to take down the runtime. Switched to `void Promise.reject(e)` so the error still surfaces (`process.on("unhandledRejection")` / `window.unhandledrejection`) but a missing handler no longer aborts the process.

3. **`packages/test/src/test/vector/{InMemoryVectorStorage,IndexedDbVectorStorage}.atomicity.test.ts` — narrow handlers.** The two atomicity test files installed file-level blanket `process.on("uncaughtException" + "unhandledRejection")` no-ops to absorb the safeEmit microtask rethrow. With fix #2 only the rejection channel is touched; replaced the blanket no-ops with describe-scoped `beforeAll`/`afterAll` handlers that only swallow the two literal `"listener boom" / "first listener boom"` messages — anything else still rethrows.

### Util / credentials

4. **`packages/util/src/credentials/CredentialStoreRegistry.ts` — fail-closed default.** The `"credential"` input resolver registered by `registerCredentialDefaults` used `resolved ?? id`, so a schema-annotated credential property whose key was never registered in the credential store silently returned the literal key name, which then flowed downstream as the credential value (e.g. as a Bearer token literal). The mcp-side `resolveAuthSecrets` already plugged this hole, but the generic input-resolver path remained silent. Now throws `MissingCredentialError` on a miss by default. Adds `setCredentialResolverStrict(value: boolean)` so legacy embedders / tests that intentionally exercise the literal-id path can opt back in. The error message contains only the lookup key (no value/secret interpolation).

## Files

- `packages/storage/src/vector/InMemoryVectorStorage.ts` (+ `packages/test/src/test/vector/InMemoryVectorStorage.atomicity.test.ts`)
- `packages/storage/src/events/safeEmit.ts`
- `packages/test/src/test/vector/InMemoryVectorStorage.atomicity.test.ts`, `packages/test/src/test/vector/IndexedDbVectorStorage.atomicity.test.ts`
- `packages/util/src/credentials/CredentialStoreRegistry.ts` (+ `packages/test/src/test/util/CredentialStore.test.ts`)
- Drive-by prettier-only cleanups picked up by `bun run format`: `packages/mcp/src/util/McpAuthProvider.ts` (import ordering), `packages/test/src/test/storage-vector/SupabaseVectorStorage.validation.test.ts` (line wrapping).

## Verification

Local results from `/home/user/libs` after the fixes:

| Command | Result |
| --- | --- |
| `bun run build:types --force` | 36/36 turbo tasks successful |
| `bun scripts/test.ts storage vitest` | 51 files passed; 1396 tests passed, 13 skipped |
| `npx vitest run packages/test/src/test/vector/` | 5 files passed; 44 tests passed (incl. 1 new "putBulk rollback does not wipe concurrent put committed mid-batch") |
| `bun scripts/test.ts rag vitest` | 24 files passed; 225 tests passed |
| `bun scripts/test.ts util vitest` | 43 files passed; 607 tests passed, 10 skipped (incl. 4 new credential resolver tests) |
| `bun scripts/test.ts mcp vitest` | 7 files passed; 109 tests passed |
| `bun run format` | clean (after the two drive-by fixes above) |

Storage suite was 1395 tests before this branch (per PR #584's report) and stays at 1396 here (the new concurrent test lives in the vector group); util was 603 before and 607 after; MCP is unchanged at 109; RAG unchanged at 225; vector grew from 43 to 44.

## Notes

- Branched from `claude/beautiful-mayer-wr1pgh` (PR #584) because the safeEmit module and atomicity test files this PR modifies are introduced there; the diff against `main` therefore includes #583's atomicity work and #584's safeEmit infrastructure. Merge #583 + #584 first.
- The `setCredentialResolverStrict(false)` opt-out is intentionally not re-exported by name in any new public barrel beyond what `@workglow/util` already exposes via the credentials module re-export — keeping it visible but not prominent.

## Test plan

- [x] `bun run build:types --force` clean across all 36 packages
- [x] Storage suite passes including the new concurrent-put-vs-putBulk-rollback test
- [x] Vector suite passes (44/44)
- [x] Util suite passes with 4 new credential resolver tests
- [x] MCP suite unchanged (the `resolveAuthSecrets` mcp path was already fail-closed; this PR fixes the orthogonal generic input-resolver path)
- [x] RAG suite passes
- [x] `bun run format` clean

https://claude.ai/code/session_01YDTm1h7PrFoLbpN9MEZX9k

---
_Generated by [Claude Code](https://claude.ai/code/session_01YDTm1h7PrFoLbpN9MEZX9k)_


sroussey:
Superseded by #589. The vector mutex and `safeEmit`→`Promise.reject` commits (plus the narrowed atomicity-test handlers) are cherry-picked there onto current `main`.

Note: the **credential fail-closed commit (`980a2ca`) was intentionally dropped** — `main` has already closed that gap. The `"credential"` input resolver no longer does `resolved ?? id`; it returns `undefined` on a miss (`resolveCredential(id, registry)`), so the literal-id leak this commit targeted no longer exists. Re-introducing the `throw MissingCredentialError` fail-closed behavior would be a separate behavior change, not a security fix, so it was left out. If you still want fail-closed semantics, that can be raised as its own PR. Closing.

---

PR #601, opened 2026-06-27
fix(task-graph): cap bridgeSubGraphTaskEvents depth to prevent event amplification

sroussey:
## Summary

`bridgeSubGraphTaskEvents` installs one re-emit listener per event type per nesting level. A deeply nested compound task (e.g. a `MapTask` containing a `GraphAsTask` containing a `MapTask` …) bridges every level independently, so a single inner `task_progress` becomes N parent emits before reaching any wire subscriber. Combined with downstream consumers that keep a bounded event log (the sec/builder `ExecutionEventLog` is capped at 10k events), sustained nested fan-out can silently evict legitimate events.

This PR adds a **defense-in-depth depth cap** with a default of **16** levels. Once exceeded, the bridge installs no listeners and emits a single `warn` log noting the cap hit. The cap is overridable via a new `maxDepth` parameter; depth is auto-derived from a symbol-keyed marker on the parent graph so existing call sites (`GraphAsTaskRunner`, `WhileTask`, `IteratorTaskRunner`, `FallbackTaskRunner`, `GraphAsTask` streaming path) compile unchanged.

### Approach

Depth tracking lives on the graph instance via a `Symbol.for("@workglow/task-graph/SubGraphEventBridge.depth")` keyed property — chosen over threading a counter through every caller because it keeps the diff to one file of fix code plus one re-export. The teardown restores the previous marker so an independently-rooted later bridge of the same subgraph instance does not inherit a stale counter.

### Files

- `packages/task-graph/src/task-graph/SubGraphEventBridge.ts` — depth cap + warn log
- `packages/task-graph/src/common.ts` — re-export `bridgeSubGraphTaskEvents` so the test can drive it directly
- `packages/test/src/test/task-graph/TaskCompleteEvent.test.ts` — three new tests

### Backwards compatibility

Patch-bump compatible: both new parameters have defaults; every existing caller still compiles. No behavior change for graphs nesting fewer than 16 bridges deep.

## Test plan

- [x] `bun scripts/test.ts graph vitest` — 74 files, 728 tests passing (including 3 new)
- [x] `bun run build:types` — 36 successful, 36 total
- [x] New tests cover: default cap kicks in at depth 16; warn log fires with `{ depth, maxDepth }` fields; `maxDepth` override is honored

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01V3e3m8cMRy5stFhDzGmZrF

---
_Generated by [Claude Code](https://claude.ai/code/session_01V3e3m8cMRy5stFhDzGmZrF)_

### DIFF
--- a/packages/task-graph/src/common.ts
+++ b/packages/task-graph/src/common.ts
@@ -16,6 +16,7 @@ export * from "./task-graph/ITaskGraph";
 export * from "./task-graph/RunContext";
 export * from "./task-graph/RunScheduler";
 export * from "./task-graph/StreamPump";
+export * from "./task-graph/SubGraphEventBridge";
 export * from "./task-graph/TaskGraph";
 export * from "./task-graph/TaskGraphEvents";
 export * from "./task-graph/TaskGraphRunner";
--- a/packages/task-graph/src/task-graph/SubGraphEventBridge.ts
+++ b/packages/task-graph/src/task-graph/SubGraphEventBridge.ts
@@ -65,7 +65,9 @@ export function bridgeSubGraphTaskEvents(
     // parentGraph is this subgraph) derives `depth >= maxDepth` and also
     // short-circuits — otherwise the depth counter resets at the next level
     // and the cap leaks downstream.
-    (subGraph as unknown as Record<symbol, number>)[BRIDGE_DEPTH] = maxDepth;
+    const subGraphAsRecord = subGraph as unknown as Record<symbol, number>;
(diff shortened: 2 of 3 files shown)


sroussey:
Superseded by #606, which cherry-picks this PR's depth-cap change onto a fresh `main` alongside #604 and applies an extra-high-effort code review on top (including a fix for an over-cap depth-marker leak in `SubGraphEventBridge` found during that review). Closing in favor of #606.

---

PR #604, opened 2026-06-28
fix(libs): janitor live-run guard, Container dispose-on-replace, vector default parity, StreamPump cancel

sroussey:
Stacks on #602. Four independent follow-up fixes.

## 1. (CRITICAL) `CacheJanitor` live-run guard
`sweepStaleRunPrivate` called `clearOlderThan` unconditionally — `createdAt` is the row write time, not run liveness, so a long-running in-flight run whose rows predate the cutoff would have its cache deleted out from under it. `CacheJanitor` now accepts a `liveRunIds: () => Iterable<string>` snapshot (default `() => []` + one-time `console.warn`) and passes the live set into `clearOlderThan(olderThanInMs, excludeRunIds)`. The repo enumerates distinct old runIds and calls `deleteRunOlderThan` per runId not in the exclude set (`ITabularStorage` has no `NOT IN` operator).

## 2. (HIGH) `Container.register` dispose-on-replace
`register` / `registerInstance` overwrote a cached singleton without invoking its disposer. A DB connection, file handle, or event subscriber re-registered would leak the original. Extract the disposer chain into `invokeDisposer` (protocol probe) + `disposeService` (eviction wrapper that catches errors with `console.warn` so a buggy disposer cannot block re-registration). `dispose()` retains its `AggregateError` collection by calling `invokeDisposer` directly. Inherited (parent-owned) instances are skipped.

## 3. (HIGH) Vector default `scoreThreshold` parity
`InMemoryVectorStorage` and `IndexedDbVectorStorage` defaulted `scoreThreshold` to `-Infinity`; every SQL backend (Sqlite / Postgres / Supabase / SqliteAi) defaults to `0`. Identical app code returned different result sets depending on the backing — a portability foot-gun. Switched in-memory + IndexedDB defaults to `0`. Callers that want every result regardless of correlation pass `scoreThreshold: -Infinity` explicitly. Inverted the two existing validation tests asserting the old behaviour.

## 4. (HIGH) `StreamPump.createStreamFromTaskEvents` cancel handler
The wired stream_chunk / stream_end / status listeners only released on a normal stream_end or terminal task status. A consumer calling `reader.cancel()` mid-stream left every listener attached: the task kept emitting into a closed controller and the listeners pinned GC. Hoist `cleanup` out of `start()` so the `cancel` handler can call it. `cleanup` is idempotent.

## Verification
- `bun run build:packages` clean (incl. types).
- `bun scripts/test.ts graph vitest` — 736 passed.
- `bun scripts/test.ts util vitest` — 648 passed, 10 skipped.
- `bun scripts/test.ts storage vitest` — 1484 passed, 13 skipped.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ERWJJbFgDbPokzJCBsFMdE)_

### DIFF
--- a/packages/indexeddb/src/job-queue/IndexedDbQueueStorage.ts
+++ b/packages/indexeddb/src/job-queue/IndexedDbQueueStorage.ts
@@ -14,7 +14,7 @@ import type {
 } from "@workglow/job-queue";
 import { JobStatus, validateLeaseMs } from "@workglow/job-queue";
 import { HybridSubscriptionManager } from "@workglow/storage";
-import { createServiceToken, deepEqual, makeFingerprint, uuid4 } from "@workglow/util";
+import { createServiceToken, deepEqual, getLogger, makeFingerprint, uuid4 } from "@workglow/util";
 import { IndexedDbMigrationRunner } from "../migrations/IndexedDbMigrationRunner";
 import { indexedDbQueueMigrations } from "../migrations/indexedDbQueueMigrations";
 import { openIdb } from "../storage/openIdb";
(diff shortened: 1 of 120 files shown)


sroussey:
Superseded by #606, which cherry-picks this PR's four fixes (janitor live-run guard, `Container` dispose-on-replace, vector default parity, `StreamPump` cancel) onto a fresh `main` — dropping the stale #602-stacked history that made this branch conflict — alongside #601, and applies an extra-high-effort code review on top (Container disposal drain-gap fix + `clearOlderThan` empty-exclude fast-path restore, both found during that review). Closing in favor of #606.

---

PR #613, opened 2026-07-06
fix(task-graph): close prefix-boundary + tier-share leaks in RunPrivateCacheRepo (follow-up to #612)

sroussey:
## Summary

PR #612 added run-scoped by-ref readers to close a cross-run leak in the FsFolder task-output cache, but left two related gaps. First, `runScopePrefix`'s `"::"` terminator is not actually unambiguous once it passes through `sanitize()` (which collapses `:` to `-`), so one runId's sanitized scope prefix can be a strict prefix of another's — a `RunPrivateCacheRepo` wrapper for one run can then read or delete blobs belonging to a different run. Second, even with that fixed, if an operator points the `CacheRegistry`'s `private` and `deterministic` slots at the same backing instance, `TaskRunner.hydrateInputRefs`'s private-then-deterministic fallback lets a foreign-run ref that the private wrapper correctly rejects still resolve through the unscoped deterministic reader — reopening the same class of leak at a different layer.

Both are closed here: an unambiguous (netstring-style) length-prefixed encoding for `runScopePrefix`, and a config-time guard in `TaskRunner` that throws when the two cache-registry slots share a backing.

## Failure scenario

Before this fix:
- `sanitize(runScopePrefix("session1"))` === `"__run-session1--"`
- `sanitize(runScopePrefix("session1-"))` === `"__run-session1---"`

The first is a strict prefix of the second, so a `RunPrivateCacheRepo` wrapper scoped to `run=session1` accepts blob names that actually belong to `run=session1-`, and can delete them via `deleteOutputByRef`.

## Approach

- **Netstring-length encoding**: `runScopePrefix(runId)` is now `` `__run:${runId.length}:${runId}::` ``. Any two distinct runIds diverge at their length digits (which sanitize to themselves) before either one's content is compared, so no runId's sanitized prefix can ever be a strict prefix of another's. This is a minimal, localized change — no schema or ref-format migration, and since PR #612's branch is unmerged, there's no back-compat burden.
- **Config-time guard**: rather than tagging every `CacheRef` with its owning tier (a larger, riskier change touching the ref schema and every reader), `TaskRunner.handleStart` now rejects a `CacheRegistry` where `private instanceof RunPrivateCacheRepo` and `private.backing === deterministic`, before any input hydration for the run. This catches the misconfiguration at its source instead of trying to make every consumer of the two slots re-derive tier ownership from ref shape.
- The **same-folder-path-but-different-instance** variant (two separate `FsFolderTaskOutputRepository`s pointed at the same directory) is a residual, currently-undetected case of the same underlying problem; documented as deferred on `CacheRegistry`.

## Test coverage

Added a `describe("prefix-boundary collision", ...)` suite in `RunPrivateFsFolderStream.test.ts` covering 3 (victim, attacker) runId pairs that reproduce distinct flavors of the pre-fix collision:
- `session1` / `session1-`
- `run-1` / `run-1:x`
- `x` / `x-y`

For each pair: the victim's wrapper cannot read (`getOutputStreamByRef` / `getOutputByRef` → `undefined`) or delete (`deleteOutputByRef` resolves as a no-op, blob count unchanged) a blob written by the attacker's wrapper, plus a positive control that the attacker can still round-trip and delete its own ref.

## Verification

```
bun scripts/test.ts graph vitest
```
1036 tests passed / 114 files passed (up from 1033 before this branch's test additions — net +3 for the new prefix-boundary suite).

```
bunx turbo run build-types --filter=@workglow/task-graph
bunx turbo run build-types --filter=@workglow/test
```
Both succeed (task-graph + its full dependency chain, and the test package).

```
grep -rn "sanitize(runScopePrefix\|sanitize(nsPrefix" packages/task-graph/src
grep -rn "runScopePrefix" packages/task-graph/src
```
Confirms every sanitized-prefix comparison routes through the single `runScopePrefix` function in `FsFolderTaskOutputRepository.ts` — no duplicated/inlined prefix logic elsewhere.

## Sequencing

1. `fix(task-graph): use netstring-length prefix in runScopePrefix to prevent collision`
2. `test(task-graph): pin cross-run prefix-boundary rejection`
3. `fix(task-graph): fail fast when private and deterministic tiers share a backing`

---
_Generated by [Claude Code](https://claude.ai/code/session_01H4LhPHFo9KQdKaEAuABcYd)_

### DIFF
--- a/packages/task-graph/src/cache/CacheRegistry.ts
+++ b/packages/task-graph/src/cache/CacheRegistry.ts
@@ -19,6 +19,21 @@ import type { TaskOutputRepository } from "../storage/TaskOutputRepository";
  *
  * Both slots are optional. When a slot is unset, caching for matching tasks is
  * a silent no-op — the task still runs correctly, just uncached.
+ *
+ * The two slots MUST NOT share a backing repository instance. `private` scopes
+ * every read/delete by `runId` (via `RunPrivateCacheRepo`'s `*ForRun` routing)
+ * specifically so one run cannot see another's rows or blobs; `deterministic`
+ * has no such scoping by design. If both slots resolve to the same backing,
+ * `TaskRunner.hydrateInputRefs`'s fallback loop — which tries `private` first
+ * and falls through to `deterministic` on a miss — lets a foreign-run ref that
+ * `private` correctly rejects still resolve via the unscoped `deterministic`
+ * reader, silently reopening the cross-run leak the scoping exists to close.
+ * `TaskRunner` throws a `TaskConfigurationError` at run-start when it detects
+ * this (`private instanceof RunPrivateCacheRepo && private.backing ===
+ * deterministic`). Note this only catches the *same-instance* case: two
+ * separate `FsFolderTaskOutputRepository` instances pointed at the same folder
+ * path are a residual, currently-undetected variant of the same
+ * misconfiguration — deferred.
  */
 export interface CacheRegistry {
   deterministic?: TaskOutputRepository;
--- a/packages/task-graph/src/cache/RunPrivateCacheRepo.ts
+++ b/packages/task-graph/src/cache/RunPrivateCacheRepo.ts
@@ -38,13 +38,13 @@ export interface RunPrivateCacheRepoOptions {
 export class RunPrivateCacheRepo extends TaskOutputRepository {
   private static fallbackWarned = false;
 
-  private readonly backing: TaskOutputRepository;
+  private readonly _backing: TaskOutputRepository;
   private readonly runId: string;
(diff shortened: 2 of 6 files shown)


sroussey:
Closing as subsumed by #615. This PR's two fixes — netstring `runScopePrefix` (prefix-boundary collision) and the same-backing `private`/`deterministic` tier guard — are identical to Fix 1 and Fix 2 in #615, which also carries three additional HIGH fixes (schema-gated portless `outputStream` discovery, `BinaryStreamRouter` push/end race, passthrough-edge watchdog). #615 has now been merged into the #608 branch (`claude/relaxed-faraday-ql3olr`), so these fixes are already landed there.

---

PR #632, opened 2026-07-13
fix(node-llama-cpp): close sequence-reclaim race and auto-evict on VRAM pressure

sroussey:
Two robustness fixes for the local GGUF provider, surfaced while running multi-model extraction sweeps:

- "No sequences left": LlamaContextSequence.dispose() reclaims its sequence id asynchronously (under the context lock), so a fast serial re-acquire on a cached single-sequence context could observe an empty pool and throw even though the prior sequence was disposed. It surfaced on slow models but not fast ones -- a race, not a leak. Add acquireContextSequence(), which waits for sequencesLeft > 0 before getSequence(), and route all six run functions through it. Waiting (rather than growing the pool) avoids extra per-sequence KV memory cost on VRAM-constrained hosts.

- VRAM exhaustion across models: the worker caches every loaded model/context without eviction, so sweeping several models in one process runs out of VRAM (each loads fine in isolation). Wrap model load and context creation in withVramEviction(), which on an allocation error evicts the least-recently- used cached model (tracked via touchModel) and retries until it fits or there is nothing left to free.

Adds a regression test that reproduces the reclaim race.

### DIFF
--- a/packages/test/src/test/ai-provider-nodellama/LlamaCpp_Runtime.test.ts
+++ b/packages/test/src/test/ai-provider-nodellama/LlamaCpp_Runtime.test.ts
@@ -4,18 +4,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { describe, expect, it } from "bun:test";
+import { acquireContextSequence } from "@workglow/node-llama-cpp/ai-runtime";
 import type { LlamaContext, LlamaContextSequence } from "node-llama-cpp";
-import { acquireContextSequence } from "./LlamaCpp_Runtime";
+import { describe, expect, it } from "vitest";
 
-/**
- * A minimal stand-in for `LlamaContext` that models node-llama-cpp's
- * **asynchronous** sequence reclamation: `sequencesLeft` reports 0 while a
- * previously disposed sequence's id is still being pushed back into the pool
- * under the context lock, then flips positive after `reclaimAfter` macrotasks.
- * `getSequence()` throws "No sequences left" whenever it is called while the
- * pool is empty — exactly as the real context does.
- */
 function makeFakeContext(reclaimAfter: number): {
   context: LlamaContext;
   getSequenceCalls: () => number;
@@ -25,9 +17,6 @@ function makeFakeContext(reclaimAfter: number): {
   const tick = (): void => {
     ticks += 1;
   };
-  // Advance a macrotask counter on every event-loop turn so `sequencesLeft`
-  // becomes positive only after `reclaimAfter` yields, mimicking the deferred
-  // lock-guarded reclaim.
   const pump = (): void => {
     if (ticks < reclaimAfter) setTimeout(pump, 0);
   };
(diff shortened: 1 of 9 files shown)


sroussey:
Follow-up commit `9c0bc26` pushed to this branch addressing three HIGH-severity issues surfaced by a follow-up review of the PR:

**H1 — eviction disposes embedding context.** Regression fix in the LRU eviction path. `recycleLlamaCppTextContext({ reloadModel: true })` now disposes any cached `LlamaEmbeddingContext` for the same model key BEFORE disposing the model. Previously `evictLeastRecentlyUsedModel` freed the `LlamaModel` while leaving a dangling `LlamaEmbeddingContext` in the cache — the next embedding request would touch a context whose backing model had been disposed. JSDoc on `recycleLlamaCppTextContext` updated accordingly.

**H2 — `getOrCreateEmbeddingContext` retries under VRAM pressure via `withVramEviction`.** Embedding-context allocation is a KV-cache allocation and can fail with the same VRAM errors as text-context and model loading. Wrapping it in `withVramEviction` matches the behavior of `getOrCreateTextContext` and `getOrLoadModel` so the first embedding after a heavy text run doesn't fail when another cached model can be evicted to reclaim VRAM.

**H3 — `isVramError` broadened from a two-gate design to a flat OR.** The previous gate (`"vram" | "cuda" | "gpu"` AND allocation phrase) rejected messages that named the backend without one of those three tokens. Now catches:
- Metal — `MTLBuffer allocation failed`, `metal allocation`
- ROCm/HIP — `hipMalloc failed`, `HIP out of memory`
- Vulkan — `VK_ERROR_OUT_OF_DEVICE_MEMORY`, `VkDeviceMemory`
- Previously-missed CUDA wording — `ggml_backend_cuda_buffer_type_alloc_buffer: allocating 8192.00 MiB on device 0: cudaMalloc failed`, `cublas`, `ggml_cuda`, `allocating buffer`, `allocation failed`

`isVramError` is now exported (`@internal`) so it can be unit-tested from `@workglow/test` — it reaches through `@workglow/node-llama-cpp/ai-runtime` via the existing `export *`.

**New tests** in `packages/test/src/test/ai-provider-nodellama/LlamaCpp_Runtime.test.ts`:
- `isVramError classification` — `it.each` over 13 rows covering positive CUDA/Metal/ROCm/Vulkan cases and negative rows (`ENOENT`, `invalid model file`, `No sequences left`, `""`); asserts both the `Error` and raw string forms.
- `recycleLlamaCppTextContext with reloadModel disposes embedding context` — seeds two model keys in all three maps and asserts that recycling `/tmp/a.gguf` clears it from all three (including `llamaCppEmbeddingContexts` — the regression guard) while leaving `/tmp/b.gguf` intact, with each `/tmp/a.gguf` dispose called exactly once.
- `getOrCreateEmbeddingContext under VRAM pressure` — first `createEmbeddingContext` call throws `CUDA error: out of memory`, second resolves; asserts the LRU model is evicted, the target context is cached, and `createEmbeddingContext` is called exactly twice.

**Verification:**
- `bun scripts/test.ts provider-nodellama vitest` — `Test Files 6 passed (6)`, `Tests 48 passed | 4 skipped (52)`.
- `npx vitest run packages/test/src/test/ai-provider-nodellama/LlamaCpp_Runtime.test.ts` — 17 passed (2 pre-existing + 15 new).
- `bun run build:types` — 38/38 tasks successful.

Commit: `9c0bc26`

---

PR #641, opened 2026-07-17
Add and harden AI provider cache checkpoints

sroussey:
## Summary
- add prompt-prefix cache checkpoint support across the AI framework and providers
- preserve caller system prompts in checkpoint-seeded local chats
- add checkpoint consumption and emission to llama.cpp tool calling
- validate Gemini cached tool declarations before explicit cache reuse
- dispose Gemini CachedContent through the owning worker runtime
- run Gemini request tests against source-level mock seams and skip only the Bun-specific worker integration under Node/Vitest

## Verification
- `bun x vitest run packages/test/src/test/ai-provider/OpenAIGeminiCheckpointParams.test.ts packages/test/src/test/ai-provider/Gemini_SessionDispose.test.ts` — 18 passed, 1 Bun-only test skipped
- `bun test packages/test/src/test/ai-provider/OpenAIGeminiCheckpointParams.test.ts packages/test/src/test/ai-provider/Gemini_SessionDispose.test.ts` — 19 passed
- `bun run build:types` — 40 packages passed

<div><a href="https://cursor.com/agents/bc-019f6cb7-7e0a-7240-8b50-4e483a817ea7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-019f6cb7-7e0a-7240-8b50-4e483a817ea7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


sroussey:
Rebased this branch onto current `main` (`f93be5f1`, which now includes #680). Force-pushed `360af23a` → `f4834c7a`; 39 commits replayed plus one fix described below.

## Conflicts resolved (3)

- **`scripts/typecheck-budget.json`** — both sides added different adjacent keys to the alphabetical map (`providers/deepseek` from main, `providers/duckdb` here). Kept both.
- **`packages/test/src/test/ai-provider-nodellama/LlamaCpp_Runtime.test.ts`** — both sides appended a different `describe` block (`withSessionLock` on main, `stealLlamaCppSession` here). Kept both suites.
- **`packages/test/src/test/ai-provider/Gemini_CachedContentFallback.test.ts`** — add/add. `main` held a `describe.skip` placeholder whose comment explicitly says the real suite "only exists on the `ai-provider-cache-checkpoints` feature branch". Took this branch's real implementation, which is what that placeholder was waiting for.

`packages/ai/src/task/base/AiTask.ts` and `AiChatWithKbTask.ts` auto-merged — #680's hunks and this branch's `getJobInput` changes don't overlap.

## One fix required by the rebase

`main` has upgraded to **vitest 4**, which takes a single function-type parameter instead of the `[args], return` pair. Two `vi.fn<[…], …>()` calls in `Gemini_CachedContentFallback.test.ts` no longer compiled; migrated to `vi.fn<(arg: T) => R>()`. Committed separately as `fix(test): migrate vi.fn generics to the vitest 4 signature after rebase on main`.

## ⚠️ Pre-existing test failure — not introduced by the rebase

`packages/test/src/test/ai-provider/OpenAIGeminiCheckpointParams.test.ts` → *"evicts and retries inline once on a NOT_FOUND"* fails:

```
Error: cache not found
  ❯ Object.generateContentStream …/OpenAIGeminiCheckpointParams.test.ts:753:31
  ❯ generateGeminiStreamWithCacheFallback providers/google-gemini/dist/ai.js:434:18
```

with the runtime logging `Gemini stream failed on cached-content request; not a CachedContent NOT_FOUND, propagating { status: 404, code: 'NOT_FOUND' }`.

This is an internal inconsistency that already existed on `360af23a`: commit `3206e7de` ("tighten reactive CachedContent NOT_FOUND matcher") changed only `Gemini_CachedContentFallback.{ts,test.ts}` and `index.ts` — it did **not** update `OpenAIGeminiCheckpointParams.test.ts`, whose last change was the older `1124619c`. That older test throws a bare `{ status: 404, code: "NOT_FOUND", message: "cache not found" }`, which the tightened matcher now deliberately rejects (its own new tests treat exactly that unscoped shape as a negative, to avoid mistaking a model 404 for a cache 404).

I have **not** changed it, because which side is correct is your call: either the old test's fixture should become cache-scoped (e.g. a `cachedContents/…` mention), or the matcher was tightened further than intended. Happy to apply whichever you prefer.

Everything else is green on the rebased branch: `bun run build:types` 41/41, `bun run build` 84/84, and `ai` + `provider` + `provider-nodellama` unit suites 625 passed / 1 failed (the above) across 63 files.

---

sroussey:
## Let the retry outcome decide whether to evict a CachedContent entry

Pushed `34f99aca` onto this branch.

### The asymmetry

On a recognized CachedContent `NOT_FOUND`, `generateGeminiStreamWithCacheFallback` evicted the shared store entry **and then** retried the request inline without the handle. Eviction-on-suspicion is the defect, and the helper's own docstring already spelled out why: the two possible mistakes are not symmetric.

- If the match is a **true** positive, evicting is right — but evicting *early* buys nothing that evicting *after* the retry wouldn't.
- If the match is a **false** positive, evicting destroyed a still-valid `CachedContent` entry. Every *other* consumer of that checkpoint then silently paid a full re-encode on its next call — and the retry that eviction was in service of was doomed to fail identically anyway.

So the old code paid an unbounded, shared, silent cost to hedge against a failure it could simply have observed.

### The new rule: the retry outcome is the evidence

"Retry without the handle" and "evict the shared entry" are now decoupled, and nothing about the entry is guessed up front:

1. On a recognized cache-scoped `NOT_FOUND`, rebuild the request **without** the cache handle and retry inline — the entry is left in place.
2. **Retry succeeds** → the handle really was the problem. Evict it now via `deleteGeminiCachedContent`, so later requests stop paying the doomed first call.
3. **Retry also fails** → the handle was *not* the problem. The entry is left intact for its other consumers.

A false positive now costs exactly one extra API call and never destroys valid shared state.

**Which error propagates when the retry also fails:** the **retry** error, not the original `NOT_FOUND`. The cache-free request is the terminal, cache-independent failure and it describes the request the caller actually wants to succeed; propagating the original would send them chasing a phantom cache problem when the real fault lies elsewhere. The original's `status`/`code` is preserved in a debug line so the discarded signal stays diagnosable. The reasoning is documented in a comment at the throw site.

`isGeminiCachedContentNotFoundError` is **unchanged** — not loosened, no negative cases removed. False positives are cheap now, so the matcher *could* be relaxed; keeping it tight means a common non-cache error (model-not-found, missing File part, 404 on an unrelated URL) still costs one call rather than two. Its docstring was updated to state that as the reason for the tightness, since the old "or you destroy the entry" justification no longer applies.

### The previously-failing test is now green

`OpenAIGeminiCheckpointParams.test.ts` → *"evicts and retries inline once on a NOT_FOUND"* (tool-use) was failing on this branch and was the sole red test. Its fixture threw a bare `Object.assign(new Error("cache not found"), { status: 404, code: "NOT_FOUND" })`. That message carries no `cachedContent` mention, so the tightened matcher deliberately does not recognize it — correct behavior, stale fixture. Baseline confirmed before touching anything:

```
 ❯ packages/test/src/test/ai-provider/OpenAIGeminiCheckpointParams.test.ts (24 tests | 1 failed) 36ms
     × evicts and retries inline once on a NOT_FOUND 12ms (retry x1)

 FAIL  ... > Gemini cachedContent NOT_FOUND fallback (tool-use) > evicts and retries inline once on a NOT_FOUND
Error: cache not found
 ❯ runStream providers/google-gemini/dist/ai.js:628:39
 ❯ generateGeminiStreamWithCacheFallback providers/google-gemini/dist/ai.js:434:18

 Test Files  1 failed | 1 passed (2)
      Tests  1 failed | 42 passed (43)
```

It was rewritten to throw a **cache-scoped** error (`cachedContents/tool NOT_FOUND`) so the matcher recognizes it, and renamed to *"retries inline once on a cache-scoped NOT_FOUND and evicts after the retry succeeds"* — the old name described an ordering that no longer exists. The bare-404 shape it used to throw was not discarded; it became a new negative case asserting that an unscoped `NOT_FOUND` is neither retried nor evicted.

Both suites now cover all three outcomes (retry succeeds → evicted; retry also fails → entry kept + retry error propagates; non-cache error → neither retried nor evicted), and assert the *ordering* — the entry is still present while the retry is issued, proving eviction is a consequence of the retry rather than a precondition for it.

**Test counts:** the two suites went from 43 tests (1 failing) to **46 tests, all passing** — `OpenAIGeminiCheckpointParams` 24 → 26, `Gemini_CachedContentFallback` 19 → 20.

### Verification

`bun run build`:

```
Cached:    84 cached, 84 total
  Time:    274ms >>> FULL TURBO
```

`bun run build:types`:

```
Cached:    41 cached, 41 total
  Time:    99ms >>> FULL TURBO
```

(Type-checking was additionally force-executed to defeat the cache after the test-file edits — `npx turbo run build-types --filter=@workglow/test --force` → `37 successful, 37 total / Cached: 0 cached`.)

`bun scripts/test.ts ai provider provider-nodellama vitest unit`:

```
Running unit tests in sections [ai+provider+provider-nodellama] — 63 file(s)

 RUN  v4.1.10

 Test Files  63 passed (63)
      Tests  629 passed | 1 skipped (630)
   Duration  64.34s
```

All three were re-run from the committed state after lint-staged. Unit kind only — no integration suites were run.

One environment note: this container has Node v22 available (no v24). CLAUDE.md's warning is specifically about Node 20's `better-sqlite3` ABI mismatch, so v22 is clear of it, but the suite has not been exercised on v24 here.

---

PR #655, opened 2026-07-22
fix(sqlite): serialize putBulk across shared-connection SqliteTabularStorage instances

sroussey:
## Summary

HIGH-priority correctness fix from a scheduled review of the last 24h of `main` (surfaced by review of the `putBulk` engine refactor in `chore: release 0.3.27`).

## The bug

`SqliteTabularStorage._putBulkInternal` guarded on better-sqlite3's driver-level `db.inTransaction`, which is scoped to the **connection**, not to the storage instance. When two `SqliteTabularStorage` instances share one `Sqlite.Database` (the normal case for repos on the same file), a second `putBulk` starting during another's inter-chunk `await` gap read `inTransaction === true`, skipped its own `BEGIN`/`COMMIT`, executed rows inside the OTHER instance's transaction, emitted `put` events, and returned success. If the first instance later rolled back on a unique-constraint violation, the second's rows silently disappeared even though its caller had received the returned entities and subscribers had already seen `put` events for phantom rows.

## The fix

- Module-scope `WeakMap<Sqlite.Database, Promise<void>>` plus a protected `connectionMutex()` helper — GC-safe, keyed by the DB handle.
- `_putBulkInternal` keys off `this.inTransaction` (per-instance) and wraps its `BEGIN`/`runBulkPut`/`COMMIT`|`ROLLBACK` bracket in `connectionMutex(...)` when not already in a transaction.
- `withTransaction` wraps its own bracket the same way, inside the existing per-instance mutex — instance-first / connection-second so two instances on one handle cannot deadlock.
- `createTxView` reports `inTransaction === true` through the proxy so the `tx`-routed `putBulk` path takes the nested branch and skips both `BEGIN` and the connection lock (the outer `withTransaction` already holds both).
- `SqliteAiVectorStorage`'s overridden `_putBulkInternal` mirrors the parent — keys off `this.inTransaction` and takes `connectionMutex` on the owned-transaction path.

## Verification

Regression tests added under `packages/test/src/test/storage-tabular/SqliteTabularStorage.integration.test.ts`:

- Two `SqliteTabularStorage` instances on DIFFERENT tables sharing one `Sqlite.Database`.
- Race `Promise.allSettled([repoA.putBulk(large), repoB.putBulk(small)])` with `runBulkPut` spied to yield microtasks and then throw on A.
- Asserts A rejects, B fulfills, and repoB's rows survive.
- Second test asserts B's `put` event stream never includes rows that were rolled back with A.

Run:
```sh
bun x vitest run packages/test/src/test/storage-tabular/SqliteTabularStorage.integration.test.ts
```

## Rollback

Revert the single file (`providers/sqlite/src/storage/SqliteTabularStorage.ts`, plus the mirror in `SqliteAiVectorStorage.ts`, plus the test). No schema/migration/API changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qb6ESTZqJBiW79n6vh1voz)_

### DIFF
--- a/packages/test/src/test/storage-tabular/SqliteTabularStorage.integration.test.ts
+++ b/packages/test/src/test/storage-tabular/SqliteTabularStorage.integration.test.ts
@@ -377,3 +377,135 @@ describe("SqliteTabularStorage regressions", () => {
     expect(await storage.size()).toBe(1);
   });
 });
+
+describe("SqliteTabularStorage putBulk across shared connection", () => {
+  const SharedSchema = {
+    type: "object",
+    properties: {
+      id: { type: "integer" },
+      val: { type: "string" },
+    },
+    required: ["id", "val"],
+    additionalProperties: false,
+  } as const satisfies DataPortSchemaObject;
+
+  async function makeTwoOnSharedConnection() {
+    await Sqlite.init();
+    const db = new Sqlite.Database(":memory:");
+    const tableA = `A_${uuid4().replace(/-/g, "_")}`;
+    const tableB = `B_${uuid4().replace(/-/g, "_")}`;
+    const repoA = new SqliteTabularStorage<typeof SharedSchema, readonly ["id"]>(
+      db,
+      tableA,
+      SharedSchema,
+      ["id"] as const
+    );
+    const repoB = new SqliteTabularStorage<typeof SharedSchema, readonly ["id"]>(
+      db,
+      tableB,
+      SharedSchema,
+      ["id"] as const
+    );
+    await repoA.setupDatabase();
+    await repoB.setupDatabase();
+    return { db, repoA, repoB };
+  }
+
+  it("sibling putBulk survives when the other rolls back on the shared connection", async () => {
+    const { repoA, repoB } = await makeTwoOnSharedConnection();
+
+    // Multi-chunk batch (SQLite bulk dialect chunks by 900 params ÷ per-row
+    // params — with 2 columns per row, 2000 rows crosses at least 4 chunks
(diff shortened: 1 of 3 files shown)


sroussey:
Closing as **superseded by #657 / #658** (both now merged to `main`).

This PR fixes a real bug — `_putBulkInternal` guarding on better-sqlite3's connection-scoped `db.inTransaction` instead of the per-instance flag, so a sibling `putBulk` on a shared `Database` handle could interleave inside another instance's transaction. #657 fixes the **same** underlying bug (its description calls out "an existing latent bug where `Sqlite.Database`'s `inTransaction` check missed the wrapper — the code now consults our own `this.inTransaction` flag as well") but generalizes the remedy into a module-level `ConnectionMutex` (`packages/storage/src/tabular/ConnectionMutex.ts`) keyed on the connection handle, covering SQLite, `SqliteAiVectorStorage`, and the PGlite Postgres path in one place. #658 then hardens it for the browser ALS shim and adds an actionable `ConnectionReentryError`.

Because both PRs rewrite the same files (`SqliteTabularStorage.ts`, `SqliteAiVectorStorage.ts`, and the shared integration test) with the more general approach, this branch's instance-local `WeakMap` mutex would conflict on merge and is redundant. The merged stack was verified before landing: 40/40 packages type-check and the SQLite + Postgres integration suites pass (298 passed / 2 pre-existing skips) against current `main` (which had also picked up the JSONB `PostgresTabularStorage` change).

No action needed here — the fix is in `main` via #657/#658.

---

PR #660, opened 2026-07-30
feat(tf-mediapipe): MediaPipe GenAI LLM inference + GPU-by-default delegates

sroussey:
## Summary

Integrates `@mediapipe/tasks-genai` into `@workglow/tf-mediapipe` so the provider actually uses the dependency it has been carrying, and turns on GPU delegates by default across the provider.

**GenAI (LLM inference, browser/WebGPU):**
- Streaming `["text.generation"]` run-fn serving both `TextGenerationTask` and `AiChatTask` (unified prompt+chat via `toTextFlatMessages`), with chat templating (`gemma`, `chatml`, `none`)
- `["text.generation", "json-mode"]` prompt-based structured generation: partial-object deltas while streaming, fence-strip + truncation-recovery JSON extraction, parsed object in `finish.data.object`
- `["model.count-tokens"]` via `sizeInTokens`, plus a preview fn that answers only from an already-loaded instance and bails while a generation holds the model lock
- `ModelDownloadTask` support for `genai-text` (warm the browser cache, then close under the lock)
- `TFMP_GenaiRuntime`: one `LlmInference` per model (single-flight creation), shared high-performance WebGPU device, per-model FIFO mutex around every SDK call, lock-guarded teardown, abort → `cancelProcessing()`
- GenAI WASM CDN pinned to the installed SDK version, with a parity unit test so the pin cannot drift from the caret-ranged dependency

**GPU by default:**
- New `provider_config.gpu` flag (default **true**): vision tasks get `baseOptions.delegate: "GPU"` with a one-shot CPU fallback on creation failure; genai always requires WebGPU; text/audio stay CPU (unsupported by the web SDK)
- `getModelTask` now deep-merges caller `baseOptions` and deep-compares cached options, so delegate variants cache correctly

**Models/wiring:** genai capability inference (gemma/qwen/`.litertlm`/llm filenames), two search catalog entries — ungated **Qwen2.5 1.5B Instruct** (int8, Apache-2.0, `chatml`) as the default and **Gemma 3 1B IT** with its Hugging Face license gate documented (anonymous downloads 401) — README usage docs, and worker/inline registration incl. preview tasks.

Sampler params (`max_tokens`, `top_k`, `temperature`, `random_seed`) are fixed at model load: the shipped 0.10.29 web SDK cannot change sampler options after load (`setOptions` requires a model asset and reloads), so per-run temperature inputs are deliberately ignored and documented as such.

## Test plan

- `bun run build:packages` — 79/79 tasks green
- `bun scripts/test.ts provider vitest` — 18 files passed, 241 tests passed (5 skipped): chat-template rendering (gemma + chatml), JSON extraction, delegate resolution, capability inference, worker/inline registration parity (order-sensitive), model-search schema validity, genai lock FIFO/rejection semantics, WASM version parity
- Adversarial multi-lens code review completed; all confirmed findings fixed (single-flight creation, lock-guarded close, preview busy-bail, `clearCancelSignals` existence guard, ungated default model)

https://claude.ai/code/session_01PRs5utsw2pvZybjx7y3MP6

---
_Generated by [Claude Code](https://claude.ai/code/session_01PRs5utsw2pvZybjx7y3MP6)_

### DIFF
--- a/packages/test/src/test/ai-provider-api/Gemini_Generic.integration.test.ts
+++ b/packages/test/src/test/ai-provider-api/Gemini_Generic.integration.test.ts
@@ -20,7 +20,11 @@ import { runAiProviderConformance } from "../../contract/ai-provider/runAiProvid
 const RUN = !!process.env.GOOGLE_API_KEY || !!process.env.GEMINI_API_KEY;
 // gemini-2.5-flash was retired from the v1beta API. Exercise a current
 // "thinking" model so the thought-signature and structured-generation paths
-// are covered live.
+// are covered live. The record pins an explicit thinking_budget: 3.5-flash
+// spends ~90+ thought tokens even on trivial prompts and Gemini counts
+// thoughts against maxOutputTokens, so without the budget padding the
+// fixture's small maxTokens is consumed by thinking and the model returns
+// MAX_TOKENS with no visible text (or tool call) at all.
 const MODEL_ID = "gemini:gemini-3.5-flash";
 const EMBED_MODEL_ID = "gemini:gemini-embedding-001";
 
@@ -41,7 +45,7 @@ runAiProviderConformance({
         description: "Google Gemini 3.5 Flash",
         capabilities: ["text.generation", "text.rewriter", "text.summary", "tool-use", "json-mode"],
         provider: GOOGLE_GEMINI as typeof GOOGLE_GEMINI,
-        provider_config: { model_name: "gemini-3.5-flash" },
+        provider_config: { model_name: "gemini-3.5-flash", thinking_budget: 1024 },
         metadata: {},
       });
       await getGlobalModelRepository().addModel({
--- a/packages/test/src/test/ai-provider/TFMP_GenaiHelpers.test.ts
+++ b/packages/test/src/test/ai-provider/TFMP_GenaiHelpers.test.ts
@@ -7,7 +7,13 @@
 import { _testOnly } from "@workglow/tf-mediapipe/ai";
 import { describe, expect, it } from "vitest";
 
-const { buildGenaiPrompt, optionsMatch, resolveTfmpChatTemplate, resolveTfmpDelegate } = _testOnly;
+const {
+  buildGenaiPrompt,
+  extractJsonFromText,
+  optionsMatch,
+  resolveTfmpChatTemplate,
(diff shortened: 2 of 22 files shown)


sroussey:
CI status on `test-vitest-ai-provider-api`: the only failures are in the **live Google Gemini text conformance** tests (`Gemini_Generic.integration.test.ts` — attempt 1: tool-call accumulator; attempt 2: tool-call accumulator + text-generation smoke, with the model returning empty candidates: `text` undefined, zero tool calls). Everything else in the job is green both attempts (114–115 passed), including the Gemini *image* tests on the same API key, and all TFMP suites.

This PR's diff is confined to `providers/tf-mediapipe` and its unit tests — it doesn't touch the Gemini provider, the shared conformance assertions, or their fixtures — so this is live-API behavior (intermittent empty responses from the Gemini text model in CI), not a regression. `typecheck-budget` from the earlier run is fixed on this head (re-baselined for the new genai run-fns).

I have a re-run of the failed job armed; if Gemini keeps returning empty candidates across further attempts, the conformance suite's pinned Gemini text model needs a look (model behavior change or key quota) — that's independent of this PR.

---

sroussey:
Root cause of the persistent `Gemini_Generic` failures found and fixed in `db7900e` — correcting my earlier "transient flake" framing: it was a real, reproducible budget collision.

Measured against the live API: `gemini-3.5-flash` spends **92–96 thought tokens even on "Say hello in one short sentence."**, and Gemini counts thoughts against `maxOutputTokens`. With the conformance fixture's `maxTokens: 100`, the model intermittently finishes `MAX_TOKENS` with an empty candidate — one probe call returned `parts=0, thoughtsTokenCount=96, no text`, which is exactly the CI assertion failure (`text` undefined / zero tool calls); tool-call prompts induce longer reasoning, so that test tripped on every CI attempt. Structured generation never failed because it already applies a default thinking budget via `resolveThinkingConfig`.

Fix: the conformance model record now pins `thinking_budget: 1024`, which sends `thinkingConfig` and pads the output cap (`maxOutputTokens: 1124`) — reasoning stays bounded, the visible answer always has room, and the thought-signature paths this model was chosen to exercise remain covered. Verified live locally: 14/14 passing across repeated runs.

---

PR #663, opened 2026-07-31
Add DeepSeek provider for @workglow/ai

sroussey:
Adds a complete DeepSeek provider implementation for the Workglow AI framework, enabling integration with DeepSeek's OpenAI-compatible chat API.

## Summary

This PR introduces `@workglow/deepseek`, a new provider package that integrates DeepSeek's language models into the Workglow AI task system. DeepSeek exposes an OpenAI-compatible REST API, so the implementation reuses the `openai` SDK pointed at `https://api.deepseek.com`.

## Key Changes

- **New provider package** (`providers/deepseek/`) with full capability support:
  - Text generation (`text.generation`)
  - Tool calling (`tool-use`)
  - Structured/JSON output (`json-mode`)
  - Text rewriting (`text.rewriter`)
  - Text summarization (`text.summary`)
  - Token counting (`model.count-tokens`)
  - Model search and info (`model.search`, `model.info`)

- **Core implementation files**:
  - `DeepSeek_Client.ts` — OpenAI SDK initialization with API key resolution and base URL validation
  - `DeepSeek_TextGeneration.ts` — unified streaming run-fn for both chat and prompt-based text generation
  - `DeepSeek_ToolCalling.ts` — tool-use support with OpenAI-compatible tool definitions
  - `DeepSeek_StructuredGeneration.ts` — JSON schema mode for structured outputs
  - `DeepSeek_CountTokens.ts` / `.browser.ts` — token counting via tiktoken (with fallback to cl100k_base encoding)
  - `DeepSeek_TextRewriter.ts`, `DeepSeek_TextSummary.ts` — specialized text tasks
  - `DeepSeek_ModelSearch.ts` — model discovery via API or fallback list
  - `DeepSeek_Capabilities.ts` — capability inference for DeepSeek model IDs

- **Provider registration**:
  - `DeepSeekQueuedProvider` — main-thread shell supporting both inline and worker-backed modes
  - `DeepSeekProvider` — worker-side provider implementation
  - `registerDeepSeek()`, `registerDeepSeekInline()`, `registerDeepSeekWorker()` — registration entry points
  - Browser and Node.js variants (token counting uses `js-tiktoken` in browser, WASM `tiktoken` in Node)

- **Model schema** (`DeepSeek_ModelSchema.ts`):
  - Discriminator: `provider: "DEEPSEEK"`
  - Config: `model_name`, optional `credential_key`, `base_url`, `trustedBaseUrl`
  - Validates base URL against allow-list (`api.deepseek.com`) unless explicitly trusted

- **Capability inference**:
  - Pattern-matches model IDs (e.g., `deepseek-v4-flash`, `deepseek-v4-pro`) to inferred capability sets
  - Falls back to declared capabilities or baseline (`model.search`, `model.info`)
  - DeepSeek chat models are text-only (no vision input, no image generation)

- **Integration**:
  - Updated `@workglow/workglow` meta-package with DeepSeek exports
  - Added DeepSeek to example projects (`eval`, `cli`)
  - Comprehensive test suite (`DeepSeekProvider.test.ts`) covering capability inference, run-fn parity, and model metadata

- **Documentation**:
  - `README.md` with feature overview, installation, and usage examples

## Implementation Details

- **OpenAI SDK reuse**: DeepSeek's API is wire-compatible with OpenAI's chat completions endpoint, so the provider leverages the existing `openai` package rather than a custom HTTP client.
- **Streaming architecture**: All run-fns use streaming (even one-shot tasks like token counting) to maintain consistency with the Workglow event-driven model.
- **Token counting approximation**: DeepSeek doesn't ship a tokenizer package, so counts are approximated using tiktoken's `cl100k_base` encoding (

https://claude.ai/code/session_01LpPXCh7MGan15jGgvJEnb2

### DIFF
--- a/.secrets/README.md
+++ b/.secrets/README.md
@@ -69,13 +69,16 @@ as separate CI secrets.
 
 ## Known credential keys
 
-| Credential key      | Env var hydrated     |
-| ------------------- | -------------------- |
-| `anthropic-api-key` | `ANTHROPIC_API_KEY`  |
-| `openai-api-key`    | `OPENAI_API_KEY`     |
-| `google-api-key`    | `GOOGLE_API_KEY`     |
-| `gemini-api-key`    | `GEMINI_API_KEY`     |
-| `hf-token`          | `HF_TOKEN`           |
+| Credential key       | Env var hydrated     |
+| -------------------- | -------------------- |
+| `anthropic-api-key`  | `ANTHROPIC_API_KEY`  |
+| `openai-api-key`     | `OPENAI_API_KEY`     |
+| `google-api-key`     | `GOOGLE_API_KEY`     |
+| `gemini-api-key`     | `GEMINI_API_KEY`     |
+| `hf-token`           | `HF_TOKEN`           |
+| `xai-api-key`        | `XAI_API_KEY`        |
+| `openrouter-api-key` | `OPENROUTER_API_KEY` |
+| `deepseek-api-key`   | `DEEPSEEK_API_KEY`   |
 
 To add another, edit `CREDENTIAL_TO_ENV` in `scripts/lib/test-credentials.ts`.
 
--- a/.secrets/credentials/8ba7cae724a568336246a8c2484316f6a7dfcddbe53a87795e696d3c7764a331.json
+++ b/.secrets/credentials/8ba7cae724a568336246a8c2484316f6a7dfcddbe53a87795e696d3c7764a331.json
@@ -1,0 +1,1 @@
+{"key":"deepseek-api-key","value":"{\"encrypted\":\"phxHdLmM959fXANdtIv100VntoYkLvHa9kTy9IzjEwSFPtja1oUsaG1xK8SGTniomAre02Mw0WQSFzK0Y18i/1sJGQ==\",\"iv\":\"vBvfAWqFKXtz+ENE\",\"createdAt\":\"2026-07-31T23:57:20.103Z\",\"updatedAt\":\"2026-07-31T23:57:20.103Z\"}"}
--- a/bun.lock
+++ b/bun.lock
@@ -59,6 +59,7 @@
         "@workglow/anthropic": "workspace:*",
         "@workglow/browser-control": "workspace:*",
         "@workglow/bun-webview": "workspace:*",
+        "@workglow/deepseek": "workspace:*",
         "@workglow/google-gemini": "workspace:*",
         "@workglow/huggingface-inference": "workspace:*",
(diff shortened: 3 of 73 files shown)


sroussey:
## Live conformance test added — and it caught two real bugs

Once the DeepSeek credential landed in `.secrets/`, nothing in the suite actually consumed `DEEPSEEK_API_KEY`, so I wired the provider into the shared `runAiProviderConformance` harness (`DeepSeek_Generic.integration.test.ts`, guarded on the env var exactly like the xAI one).

It failed on the first run against the real API. **My "OpenAI wire-compatible" assumption was too broad in two places** — both hard 400s, so structured output and forced tool calls were simply broken as originally submitted:

**1. `json-mode` — `400 This response_format type is unavailable now`**

DeepSeek does not accept OpenAI's `response_format: {type: "json_schema"}`. It supports only `{type: "json_object"}`, which does **no schema enforcement** and additionally requires the word "json" to appear in the prompt or the response can come back empty. The schema now travels in the prompt instead.

Worth flagging for anyone using this for extraction: `StructuredGenerationTask` still re-validates the parsed object against the output schema, so a model that ignores the shape fails loudly rather than silently returning the wrong thing — but this is genuinely weaker than the Anthropic/OpenAI/Gemini (server-enforced) and llama.cpp (grammar-constrained) paths. I noted that caveat in sec's docs where DeepSeek is proposed as a cheap extraction candidate.

**2. `tool-use` — `400 Thinking mode does not support this tool_choice`**

The v4 models are thinking models and accept only `auto` / `none`. The shared `mapOpenAIToolChoice` emits `"required"` and the named-function object, so it can't be used here. A forcing choice is now downgraded to `auto` (request succeeds, and the model still calls the tool when the prompt calls for one); `none` passes through. The tradeoff: **a caller passing `"required"` does not get a hard guarantee of a tool call from this provider** — that's a vendor limitation, not something the adapter can paper over honestly.

Text generation, streaming, mid-stream abort, and multi-turn tool use were already correct and needed no change.

**After the fix:** `DeepSeek_Generic.integration.test.ts` 14 passed / 5 skipped (skips are embeddings + sessions, which the provider correctly declares `false`). Full `ai-provider-api` suite 464 passed / 60 skipped, and the broader run 1101 passed, both exit 0.

One note on that broader run: an earlier pass showed a single failure in a conformance `beforeAll` timeout, which did not reproduce across two subsequent full runs — a flake under concurrent live-API load, not related to this diff. Worth knowing that now that several live provider suites run with real credentials, that hook is a plausible source of intermittent CI red.

---

PR #667, opened 2026-08-02
Add disown() and setTitle() for task lifecycle management

sroussey:
## Summary

Adds two new capabilities to task execution: `disown()` for releasing owned child tasks from a parent's subgraph, and `setTitle()` for relabeling task instances that are reused across multiple jobs.

## Key Changes

- **`disown()` method** — New `IExecuteContext.disown()` allows tasks to release previously owned children, keeping the subgraph flat when iterating over unbounded worklists. Uses a `WeakMap` to track wrapper tasks created by `own()`, enabling proper cleanup without memory leaks.

- **`setTitle()` method** — New `Task.setTitle()` allows relabeling task instances at runtime. Essential for reused task instances that process multiple jobs sequentially, so progress UIs display the current job name rather than the first one.

- **Task removal event** — Added `task_removed` event to `TaskGraph` (emitted by `disown()`) so the CLI UI can drop stale rows and unwire task IDs for re-registration.

- **CLI UI updates** — Updated `taskGraphCliSubscriptions.ts` to listen for `task_removed` events and clean up task info. Updated `cliTaskUi.ts` to re-read task labels on each poll, reflecting dynamic title changes.

- **Context propagation** — Added `disown` to all `IExecuteContext` implementations: `TaskRunner`, `WhileTaskRunner`, `StreamProcessor`, and test mocks.

## Implementation Details

- `disown()` is a no-op for values never owned by the context, making it safe to call defensively
- `own()` returns the original value (not the wrapper task), so `disown()` must map through `ownedWrappers` to find the actual task to remove
- `WeakMap` ensures owned tasks that are never disowned still die with the subgraph between graph runs
- `setTitle()` updates `config.title`, which is re-read by `task.title` getter on each access
- Tests added for all three scenarios: looping with disown, workflow resolution, and stray disown calls

https://claude.ai/code/session_01XgMNjWhP23c83zZGeZua4U

### DIFF
--- a/examples/cli/src/ui/cliTaskUi.ts
+++ b/examples/cli/src/ui/cliTaskUi.ts
@@ -7,6 +7,7 @@
 import type { TaskGraph } from "@workglow/task-graph";
 import type { Dispatch, SetStateAction } from "react";
 import type { CliTaskLine } from "./taskGraphCliSubscriptions";
+import { cliTaskLabel } from "./taskGraphCliSubscriptions";
 
 /** Single-character status column: done → active (spinner in UI) → waiting → error. */
 export function cliTaskStatusGlyph(status: string): string {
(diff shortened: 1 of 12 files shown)


sroussey:
Note on sequencing: this PR no longer has an in-tree consumer. workglow-dev/sec#222 originally used both `context.disown` and `Task.setTitle`, but its CI installs `workglow@0.3.34` from the registry and couldn't compile against them, so that PR was decoupled to ship independently.

This one still stands on its own — `disown` is the general fix for `own` being add-only, it's covered by tests in `OwnTask.test.ts` (per-iteration owner stays flat; workflow-to-wrapper resolution; no-op for a value never owned), and sec will pick both APIs back up once a release lands and its pin is bumped.

The coverage bot reports "No changed files found" because the changed packages aren't in the coverage run's scope; the new behaviour is tested.

---

PR #668, opened 2026-08-04
fix(cli): release a disowned task's listeners instead of retaining it forever

sroussey:
## What retains what, and why `disown` alone was not enough

`subscribeTaskGraphForCli` (`examples/cli/src/ui/taskGraphCliSubscriptions.ts`) retained **every task the graph ever contained**, for the whole life of the subscription:

- `iterationUnsubs` was a plain append-only `Array<() => void>`. `wire()` pushed one closure per wired task and each closure captured `task` (it calls `task.events.off(...)`). Nothing ever removed an entry.
- `onTaskRemoved` deleted from `wired` and from `setTaskInfos`, but **never touched `iterationUnsubs`** — that array was the strong reference that survived `context.disown`.
- `onTaskRemoved` also never deleted the task's `iterationSlots` key → unbounded `Map` growth.
- `registerTaskListeners` returned `void`, so its `status` / `progress` listeners could **never** be removed at all — even the teardown could not drop them.

So `context.disown` did its half of the job (the node leaves the graph) and the CLI subscription silently undid it: the task, and transitively its whole subtree, stayed reachable. This is reached in practice through `examples/cli/src/ui/rows/useSubtaskRows.ts:61`, which calls `subscribeTaskGraphForCli(task.subGraph, …)` for every row.

## The fix

1. **`registerTaskListeners` returns a disposer.** The two inline handlers are hoisted to named consts (`onStatus`, `onProgress`); the returned closure `off`s both. `TaskStatus` added to the type-only import.
2. **`wired` + `iterationUnsubs` collapse into one `Map<string, () => void>` (`taskUnsubs`)**, keyed by task id. The Map doubles as the already-wired guard, and stores a single combined disposer (task listeners + iteration listeners when `setIterationSlots` is provided).
3. **`onTaskRemoved` disposes**: looks up the disposer, calls it, deletes the Map entry, deletes the id from `setTaskInfos` **and** from `setIterationSlots` (the latter was missing). A removed id can still be re-wired if the same node is owned again later — the behavior the old `wired.delete(id)` provided is preserved by the Map delete.
4. **The teardown disposes everything**: iterates `taskUnsubs.values()`, calls each, `clear()`s, then `off`s the four graph listeners as before.

## Downstream: the sec sweep this unblocks

`workglow-dev/sec`'s unbounded backfill sweeps own one `Workflow` per filing and now release it with `context.disown`. On a TTY run (the progress UI is subscribed) the retained array pinned every one of those workflows — and each one's whole subtree, ending at a multi-MB fetched SEC document — for the life of the sweep. With this change the disowned workflow is actually collectable.

## Falsification

Two new tests in `examples/cli/src/test/taskGraphCliSubscriptions.test.ts` (`subscribeTaskGraphForCli task lifecycle`):

- **"disposes a removed task's listeners"** — builds a real `TaskGraph` + `Task`, subscribes, asserts the row and `listenerCount === 1` for `status` / `iteration_start`, emits `iteration_start` and asserts the slots populate; then `graph.removeTask(task.id)` (what `disown` triggers) and asserts the row **and** the slots key are gone and `listenerCount` is `0` for `status`, `progress`, and all three iteration events; then re-emits and asserts nothing is written back — inert, not merely detached.
- **"re-wires a task owned again after being disowned"** — remove, re-add, assert the row is back and `listenerCount("status") === 1`. This pins the behavior that collapsing `wired` into `taskUnsubs` must not lose.

Both were run against the **unpatched** source (`git stash` of the one source file) and both fail there:

```
FAIL … > disposes a removed task's listeners
  expected true to be false            ← the iterationSlots key persists
FAIL … > re-wires a task owned again after being disowned
  expected 2 to be 1                   ← re-wiring stacks a second status listener
```

With the fix, all 5 tests in the file pass.

Verified:
- `bunx vitest run` in `examples/cli` — 5 passed in the touched file; 23 passed / 1 skipped for the whole package.
- `bun scripts/test.ts task vitest` — 1126 passed / 24 skipped.
- `bun scripts/test.ts util vitest` — 668 passed / 10 skipped.
- `bun run build:types`, plus a forced `tsgo --noEmit` in `examples/cli`, `packages/util`, and `packages/test` (turbo reported a types cache hit for `@workglow/cli`, so the direct run is the real signal).

## ⚠️ Second commit is separable: `EventEmitter.on` dedupe

`8223767` (`packages/util/src/events/EventEmitter.ts`) makes `on` skip the push when an identical **non-`once`** listener reference is already registered, with tests in `packages/test/src/test/util/EventEmitter.test.ts`.

**This does not fix the CLI bug above.** `registerTaskListeners` allocates fresh closures on every call, so nothing there is reference-equal — commits 1 and 2 are independent, and commit 1 is the load-bearing fix. The dedupe is a guard for *stable-reference* call sites (`Task.subGraph`'s `_taskAddedListener`, `WorkflowEventBridge`) that today rely on hand-written off-before-on discipline; because `off` removes a single entry, a double-`on` there leaves a subscription no caller can unhook.

It is a **deliberate divergence from Node's `EventEmitter`**, which does duplicate. An `on` plus a `once` of the same function deliberately remain two subscriptions. If reviewers would rather keep Node parity, drop `8223767` — the first commit stands alone.

## Follow-up (maintainer step, not done here)

A **`0.3.37` release** of `@workglow/cli` (and `@workglow/util` if the second commit lands) is needed before `workglow-dev/sec` can depend on this fix. No release was run and nothing was published from this branch.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VcERQoSbBT3B2b7QHSbXPt)_

### DIFF
--- a/examples/cli/src/test/taskGraphCliSubscriptions.test.ts
+++ b/examples/cli/src/test/taskGraphCliSubscriptions.test.ts
@@ -4,13 +4,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { ITask } from "@workglow/task-graph";
+import { Task, TaskGraph, type ITask } from "@workglow/task-graph";
 import { EventEmitter } from "@workglow/util";
 import { describe, expect, it } from "vitest";
 import {
   FULL_SLOT_TRACKING_MAX,
   MAX_RUNNING_ROWS,
   registerIterationListeners,
+  subscribeTaskGraphForCli,
+  type CliTaskLine,
   type IterationSlotRow,
 } from "../ui/taskGraphCliSubscriptions";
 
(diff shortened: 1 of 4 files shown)


sroussey:
Closing in favour of #669, which fixes the same CLI listener retention plus two other `disown`-adjacent issues. Accounting for what happened to each commit here.

### `b840044` — CLI listener release: superseded, equivalent

#669 landed the same fix independently and structurally identically: `registerTaskListeners` returns a disposer, `wired` becomes a `Map` keyed by task id that doubles as the already-wired guard, `onTaskRemoved` disposes and clears both the row and the iteration-slot key, and the teardown disposes every task rather than only the iteration listeners. The only differences are cosmetic (a `Map<string, Array<() => void>>` instead of a single combined closure).

Two things from this PR were genuinely better and have been carried over in [`da483c7`](https://github.com/workglow-dev/libs/commit/da483c7):

- `onStatus` typed as `TaskStatus` rather than `string` — that is what `TaskEvents` actually emits.
- The stronger removal assertions: all three iteration events asserted to unsubscribe (not just `iteration_start`), and the "inert, not merely detached" check that a late event after removal writes nothing back.

#669 also covers a removal path this PR's tests do not: `regenerateGraph()`, which clears a subgraph task-by-task between runs and is how `useSubtaskRows`-subscribed subgraphs empty in the real UI.

### `8223767` — `EventEmitter.on` dedupe: validated, deliberately not carried over

Taking you at your word that this one is separable and reviewer's choice. I validated it rather than just dropping it:

- **It breaks nothing.** With the commit applied: `util` 669 passed, `task`+`graph` 1877 passed, `storage`+`queue` 2252 passed.
- **The asymmetry it targets is real.** `off` splices a single match, so a double-`on` of the same reference leaves a listener behind that off-once lifecycle code will not clear.
- **Cost is negligible at realistic listener counts**, but it is quadratic: adding N distinct listeners to one emitter is flat (~0.65 ms at any N) today versus ~35 ms at N=5000 with the `some()` scan. Irrelevant at the single/double-digit counts in this codebase; worth knowing since `maxListeners` defaults to `0` (unlimited), so nothing warns.

Not carrying it over regardless, for reasons that are about scope rather than correctness:

1. Both call sites named in the description already do off-before-on correctly — `Task.set subGraph` offs the previous graph first, and `WorkflowEventBridge.attach` calls `detach()` first. So there is no bug in tree that this fixes today.
2. It changes a foundational contract for every `@workglow/util` consumer, and deliberately diverges from Node's `EventEmitter`. That is a library-semantics decision worth making on its own merits, not as a rider on a PR about `disown` and data-URI decoding.
3. It turns a double-subscribe — normally the symptom of a lifecycle bug — into a silent no-op, removing the diagnostic.

The commit applies cleanly and is fully tested if you'd rather have the guard; say the word and I'll open it as its own PR against `main`.

### Release note carried forward

The `0.3.37` release of `@workglow/cli` that `workglow-dev/sec` needs is still outstanding, and now depends on #669 rather than this branch.

---

sroussey:
Correcting one claim in my previous comment. I said there was "no bug in tree that this fixes today". That was wrong — a sweep for stable-reference `on` sites finished after I posted, and it found one.

**`examples/web/src/graph/TaskNode.tsx`** had the same defect this PR fixes in the CLI, and one step worse: it never listened for `task_removed` at all. A subtask released with `context.disown` kept its `progress`/`status` subscriptions and its row, and the `task_added` handler wired `getTasks()[length - 1]` instead of the task that was actually added. Since re-adding a removed task moves it to the tail (verified against a real `TaskGraph`), owning a reused subtask again re-subscribed it — so an own/disown loop over N jobs left 2N subscriptions on one instance and recomputed consolidated progress once per duplicate, with the append-only `unsubscribes` list retaining every subtask the subgraph ever held until unmount.

That is precisely the shape `8223767` was guarding against, so the dedupe would have masked the duplicate-dispatch half of it.

It does not change the conclusion, though — if anything it sharpens it. The duplicate `on` was a symptom of the missing `task_removed` handling, and deduping would have hidden the symptom while leaving the retention (the part that actually matters for the `sec` sweep) in place: the subscriptions would still never have been released, because nothing was calling `off`. Fixed at the call site in #669 ([`52a484e`](https://github.com/workglow-dev/libs/commit/52a484e)) — subtask subscriptions keyed by id so removal can drop them, `task_added` wiring the emitted id, and a `task_removed` handler.

Everything else in the previous comment stands. The offer stands too: the `EventEmitter` guard is validated and applies cleanly if you want it as its own PR — just now with no known caller depending on it.

---

Issue #675, opened 2026-08-05
Credential store adapters for AWS Secrets Manager, HashiCorp Vault, and GCP Secret Manager


sroussey:
Split out of #238. This is the one part of that issue's ask that is genuinely unimplemented.

## Current state

The credential subsystem itself is complete — `ICredentialStore` and the `CREDENTIAL_STORE` token in `packages/util/src/credentials/`, with `InMemoryCredentialStore`, `EnvCredentialStore`, `ChainedCredentialStore`, and, in `packages/storage/src/credentials/`, `EncryptedKvCredentialStore`, `LazyEncryptedCredentialStore`, `ServerCredentialStore`, and a `SecretVault` OS-keychain seam.

What does not exist is any hosted-secret-manager backend. `providers/aws` today contains only SQS job-queue code; a repo-wide search for a secrets-manager, Vault, or GCP Secret Manager integration returns nothing.

## Why this was deliberately not built alongside the rest

Recording the reasoning so it isn't relitigated:

- `ICredentialStore` is a six-method interface, so each adapter is small — but each drags in a heavyweight SDK peer dependency.
- Each has a distinct auth model: IAM role / STS for AWS; token, AppRole, or Kubernetes auth for Vault; ADC or workload identity for GCP.
- None can be meaningfully tested in CI without live accounts or a large mocking layer.
- There is no consumer in this repo asking for one yet.

Doing all three wholesale was the single largest risk in the original issue, which is why it was carved out rather than attempted.

## Suggested shape when this is picked up

One adapter at a time, driven by a real use case, each as a sub-path export of its existing provider package (e.g. `providers/aws` gaining a credentials export implementing `ICredentialStore` against `@aws-sdk/client-secrets-manager`) so the SDK stays an optional peer dependency rather than a hard dependency of `util` or `storage`.

Adapters compose with `ChainedCredentialStore`, so a hosted backend can sit behind the env/in-memory stores without changing any call site.

PR #677, opened 2026-08-05
fix(tasks): keep resolved credentials out of queued job payloads, add credential schemes

### LINKED ISSUE #238: Credential / Secrets Management
Credential / Secrets Management 
need centralized encrypted credential storage with AWS/Vault/GCP integration. Workglow's ServiceRegistry has no credential concept.

Implementation ideas:

ICredentialStore interface in packages/util/src/credentials/
Implementations: InMemoryCredentialStore, EncryptedKvCredentialStore, EnvCredentialStore
Tasks access via context.registry.get(CREDENTIAL_STORE).get("my-api-key")
Update FetchUrlTask and AI tasks to support credential references

### LINKED ISSUE #673: FetchUrlTask writes the resolved credential into queue storage on the queued path
Split out of #238, which is otherwise implemented.

## The defect

`packages/tasks/src/task/FetchUrlTask.ts` carries this comment above its credential handling:

> Apply credential as Authorization header and strip it from the payload so the secret is not persisted to queue storage.

It strips the `credential_key` **field**, but then writes the plaintext secret into `headers.Authorization` on the *same* `jobInput` object, and that object is what gets handed to the queue:

```ts
const handle = await registeredQueue.client.send(jobInput as Input, { ... });
```

Queue bodies are durably persisted by the `IJobStore` / `IMessageQueue` implementations (SQLite, Postgres, SQS). So on the queued path the bearer token is written to storage, which is the exact outcome the comment claims to prevent.

## Scope

- Only affects the **opt-in** queued path (`config.queue` truthy). The default is `queue ?? false`, and the inline path is safe.
- Requires the caller to be using both a credential reference and queueing.
- Exposure is a secret at rest in the operator's own queue table, not a remotely reachable leak.

## No test coverage

`packages/test/src/test/task/FetchTask.test.ts` currently has **zero** credential coverage, which is why this went unnoticed.

## Suggested fix

Keep the resolved secret off the queued payload entirely. If a caller explicitly sets `config.queue` *and* supplies a credential, raise a configuration error rather than silently downgrading to the inline path — a silent downgrade would quietly lose rate limiting, which is usually the reason queueing was requested.

Resolving the credential inside `FetchUrlJob.execute()` is **not** a valid alternative: `.claude/CLAUDE.md` forbids run-fns from touching credential stores, since they execute in workers with a separate `globalServiceRegistry`.

The misleading comment should be corrected to state the real invariant.

## Related stale documentation

`docs/technical/15-credential-management.md` states:

> If the credential store does not contain the requested key, the resolver returns the original string unchanged.

The code deliberately returns `undefined` instead — see the comment in `packages/util/src/credentials/CredentialStoreRegistry.ts` explaining that echoing the id would thread the credential's *name* through as its *value*, straight into an `Authorization` header. The doc currently documents the vulnerability as the intended behavior and should be corrected alongside this fix.

### LINKED ISSUE #674: FetchUrlTask hard-codes `Authorization: Bearer` for credential references
Split out of #238, which is otherwise implemented.

`packages/tasks/src/task/FetchUrlTask.ts` resolves a `credential_key` (via the `format: "credential"` input resolver) and unconditionally places it as:

```
Authorization: Bearer <secret>
```

That covers OAuth-style APIs and nothing else. `X-Api-Key`-style headers, `Basic`, and custom schemes — i.e. most non-OAuth HTTP APIs — cannot be expressed at all today.

## Suggested shape

Two additional ports alongside the existing `credential_key`. Only `credential_key` carries `format: "credential"`; the other two are plain non-secret config and stay safe to serialize and persist:

- `credential_scheme`: `bearer` | `basic` | `header` | `none`, default `bearer`
- `credential_header`: string, default `Authorization`, used when the scheme is `header`

Defaulting to `bearer` keeps this a non-breaking change.

For `basic`, treat the secret as a pre-encoded `user:pass` base64 blob rather than inventing a splitting/encoding step inside the task.

## Notes

- Extract the placement logic into a small exported pure function so it can be unit-tested directly.
- Validate the header name (something like `/^[A-Za-z0-9-]{1,64}$/`) and reject otherwise — a caller-supplied header name reaching the request unchecked is a header-injection vector.
- Precedence against a user-supplied `headers.Authorization` should be decided, asserted in a test, and documented.

sroussey:
Fixes the two remaining defects in `FetchUrlTask`'s credential handling. The broader credential subsystem (`ICredentialStore`, the stores, `CREDENTIAL_STORE`, the `format: "credential"` resolver path, AI credential resolution) was already shipped and is unchanged here — see the now-closed #238.

## Security defect (#673)

`FetchUrlTask` resolved `credential_key` into the real secret and wrote it as an `Authorization` header onto **the same object** that was then handed to the queue:

```ts
jobInput = { ...rest, headers: { ...input.headers, Authorization: `Bearer ${credential}` } };
// ...
const handle = await registeredQueue.client.send(jobInput as Input, { ... });
```

A source comment claimed the credential was stripped "so the secret is not persisted to queue storage". It stripped the `credential_key` field but not the header it had just written. Queue bodies are persisted durably (SQLite / Postgres / SQS), so the bearer token was written to storage.

**Scope: the opt-in queued path only** (`config.queue` truthy). The default inline path (`queue ?? false`) never reached the queue and was not affected.

### Behavior chosen

The queued path now **refuses to run** when a credential is present, raising a `TaskConfigurationError`:

```
FetchUrlTask: credential_key cannot be combined with the queued path (config.queue),
because queued job payloads are persisted to durable storage. Remove config.queue to
run inline, or have the queue worker supply the credential itself.
```

Explicit refusal rather than a silent downgrade to the inline path — a downgrade would quietly drop the per-domain rate limiting that is the whole reason a caller asks for a queue. Resolving the credential inside `FetchUrlJob.execute()` was rejected as the alternative: run-fns execute in workers with a separate `globalServiceRegistry`, so touching a credential store there violates the worker-isolation rule in CLAUDE.md.

All three credential ports are now stripped from the job input unconditionally, so a resolved secret only ever exists as an in-process request header. The misleading comment is replaced with the real invariant.

## Credential schemes (#674)

`Authorization: Bearer` was hard-coded, so non-OAuth APIs could not use credential references. Added two plain non-secret config ports (only `credential_key` carries `format: "credential"`):

| Field | Type | Default |
|-------|------|---------|
| `credential_scheme` | `bearer \| basic \| header \| none` | `bearer` |
| `credential_header` | `string` | `Authorization` |

`bearer` remains the default, so **this is not a breaking change**. `basic` sends the secret verbatim as a pre-encoded `user:pass` base64 blob — no splitting or encoding is invented here. `header` places the raw secret in `credential_header` for API-key style services.

`credential_header` is validated as a bare header token (1–64 chars of letters, digits, hyphens); anything else raises a `TaskConfigurationError`. This rejects the CR/LF, colon and whitespace that would otherwise permit header injection. Validation runs even for schemes that ignore the field, so a malformed name is reported rather than silently discarded.

Placement logic is extracted into the exported pure function `applyCredentialToHeaders()` (`packages/tasks/src/task/FetchUrlCredentials.ts`) so it is unit-testable directly.

**Precedence decision:** a resolved credential **overwrites** a same-named header supplied in `headers`. The credential store is authoritative, and letting a hard-coded header shadow it would silently defeat the configured credential. This matches the previous behavior and is now asserted and documented.

## Docs (#673)

`docs/technical/15-credential-management.md` claimed that on a store miss "the resolver returns the original string unchanged". The code deliberately returns `undefined` — echoing the id would thread the credential's *name* through as its *value*, sending the key name to the remote host in an `Authorization` header and masking the misconfiguration. The doc now matches the code, and gains a section covering the new fields and the queue interaction.

## Tests

`packages/test/src/test/task/FetchTask.test.ts` had zero credential coverage. Added 12 tests using `InMemoryCredentialStore` on a `ServiceRegistry` scoped to its own `Container`, plus direct unit tests of the pure function.

## Verification

The G1 regression test was written first and confirmed failing on unmodified `main`. The failure shows the secret verbatim inside the persisted queue row:

```
FAIL  FetchUrlTask > credentials > refuses the queued path when a credential is present, and persists no secret
AssertionError: expected '[{"queue":"credential-leak-queue","in…' not to contain 'sk-super-secret-value-1234567890'

Received: "[{"queue":"credential-leak-queue","input":{"method":"GET",
"headers":{"Authorization":"Bearer sk-super-secret-value-1234567890"},
"response_type":null,"url":"https://api.example.com/data"},...,"status":"COMPLETED",...}]"

 ❯ packages/test/src/test/task/FetchTask.test.ts:915:31
```

After the fix:

```
$ npx vitest run packages/test/src/test/task/FetchTask.test.ts
 Test Files  1 passed (1)
      Tests  56 passed (56)

$ bun run build

$ bun scripts/test.ts task vitest
 Test Files  72 passed (72)
      Tests  1152 passed | 24 skipped (1176)

$ bun run build:types

$ npx vitest run packages/test/src/test/util/CredentialStore.test.ts packages/test/src/test/task/FetchUrlSsrf.test.ts
 Test Files  2 passed (2)
      Tests  89 passed (89)
```

## Not in scope

AWS Secrets Manager / HashiCorp Vault / GCP Secret Manager adapters are tracked separately in #675 and are not attempted here — three heavyweight SDKs with three auth models, untestable in CI, and no consumer in this repo currently needs them.

Closes #673
Closes #674

---

### DIFF
--- a/docs/technical/15-credential-management.md
+++ b/docs/technical/15-credential-management.md
@@ -75,9 +75,13 @@ The full resolution sequence during task execution:
    config.credential_key === "sk-ant-..."  // Already the real API key
 ```
 
-If the credential store does not contain the requested key, the resolver returns the
-original string unchanged, allowing `getClient()` to fall through to the `api_key`
-or environment variable checks.
+If the credential store does not contain the requested key, the resolver returns
+`undefined` — **never the reference string**. Echoing the id back would thread the
+credential's *name* through as its *value*, so a missing key would end up sent as a
+real secret (for example straight into an `Authorization: Bearer <key-name>` header),
+leaking the key name to the remote host and masking the misconfiguration. Consumers
+treat `undefined` as "no credential" and fall through to the `api_key` or environment
+variable checks, or raise their own missing-key error.
 
 ---
 
@@ -249,6 +253,64 @@ extends it with provider-specific fields:
 The `credential_key` field is marked with `"x-ui-hidden": true` in provider schemas,
 keeping it out of end-user form UIs while still participating in the automatic
 resolution system.
+
+---
+
+## Credentials in FetchUrlTask
+
+`FetchUrlTask` consumes credentials through the same `format: "credential"` path: its
+`credential_key` input is resolved to the real secret before `execute()` runs. Two
+sibling ports — plain, non-secret configuration — decide where that secret goes:
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `credential_key` | `string` (`format: "credential"`) | -- | Key to look up in the credential store. |
+| `credential_scheme` | `"bearer" \| "basic" \| "header" \| "none"` | `"bearer"` | How the resolved secret is placed on the request. |
+| `credential_header` | `string` | `"Authorization"` | Header name used when `credential_scheme` is `"header"`. |
+
+- **`bearer`** -- `Authorization: Bearer <secret>`. The default, so existing graphs
+  are unaffected.
+- **`basic`** -- `Authorization: Basic <secret>`. The secret is used **verbatim** and
+  must already be the base64 encoding of `user:pass`; the task performs no splitting
+  or encoding of its own.
+- **`header`** -- the raw secret becomes the value of `credential_header`, for
+  API-key style services (e.g. `X-Api-Key`).
+- **`none`** -- the credential is resolved but never placed on the request.
+
+`credential_header` must be 1-64 characters of letters, digits, or hyphens; anything
+else raises a `TaskConfigurationError`. The narrow allow-list rejects the CR/LF, colon
+and whitespace characters that would otherwise let a configured header name inject
+additional headers into the outbound request. Validation runs even for schemes that
+ignore the field, so a malformed name is reported rather than silently discarded.
+
+The placement logic is the exported pure function `applyCredentialToHeaders()`
+(`packages/tasks/src/task/FetchUrlCredentials.ts`), so it can be exercised directly.
+
+### Credentials and the queued path
+
+`FetchUrlTask` can dispatch through a job queue (`config.queue`) to get per-domain
+rate limiting. **A credential cannot be combined with that path.** Queue payloads are
+persisted durably (SQLite / Postgres / SQS), so a resolved secret written into the
+queued job input would be written to storage. Setting both raises a
+`TaskConfigurationError`:
+
+```ts
+// Throws: credential_key cannot be combined with config.queue
+new FetchUrlTask({ queue: "fetch:api.example.com" })
+  .run({ url: "https://api.example.com/data", credential_key: "my-api-key" });
+```
+
+The refusal is deliberate and explicit rather than a silent downgrade to the inline
+path, which would quietly drop the rate limiting the caller asked for. To use both,
+have the queue worker supply the credential itself instead of threading it through
+the job payload.
+
+The invariant holds regardless of scheme: `credential_key`, `credential_scheme`, and
+`credential_header` are all stripped from the job input, so the resolved secret only
+ever exists as an in-process request header. Note that a resolved credential
+**overwrites** a same-named header the caller supplied in `headers` -- the credential
+store is authoritative, and letting a hard-coded header shadow it would silently
+defeat the configured credential.
 
 ---
 
--- a/packages/tasks/src/common.ts
+++ b/packages/tasks/src/common.ts
@@ -21,6 +21,7 @@ export * from "./task/ArrayTask";
 export * from "./task/DateFormatTask";
 export * from "./task/DebugLogTask";
 export * from "./task/DelayTask";
+export * from "./task/FetchUrlCredentials";
 export * from "./task/FetchUrlJobError";
 export * from "./task/FetchUrlTask";
 export * from "./task/HumanApprovalTask";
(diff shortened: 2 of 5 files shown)


sroussey:
`test-vitest-ai-provider-hft` failed on `24b3093` for an external reason, not this diff.

`ZeroShotTasks.integration.test.ts` downloads model configs from the Hub at test time, and the Hub returned **HTTP 429**:

```
PermanentJobError: Provider HF_TRANSFORMERS_ONNX failed for ImageClassificationTask:
Error (429) occurred while trying to load file:
"https://huggingface.co/Xenova/clip-vit-base-patch32/resolve/main/config.json"
```

Same for `Xenova/owlvit-base-patch32` in the `ObjectDetectionTask` cases. Every failure in the job is a 429 on a Hub fetch; there are no assertion failures.

This PR touches `packages/tasks/src/task/FetchUrlTask.ts`, a new `FetchUrlCredentials.ts`, its tests, and one doc — nothing in the HFT provider or its model loading. The same shard passed on #672 and #676, which share this PR's base commit.

Likely cause is contention: several PRs from this triage pass had CI running concurrently, each pulling the same models from the Hub. I'll re-run the failed job once the run finishes (GitHub refuses a re-run while it is still in progress). If it reproduces after a clean re-run I'll investigate further rather than assume flakiness.

---

PR #678, opened 2026-08-05
feat(tasks): add webhook, Slack, and Discord notification tasks

### LINKED ISSUE #241: Webhook Notification Tasks
Why: Make easier to connect to Slack, Discord, email etc. Workglow has FetchUrlTask for HTTP but no structured integration tasks. Adding even basic notification tasks makes workflows immediately more useful for real-world automation.
What: Add WebhookNotifyTask (generic HTTP POST), SlackNotifyTask, and DiscordNotifyTask as simple, focused tasks.

Files to create:

packages/tasks/src/task/WebhookNotifyTask.ts - Generic HTTP webhook notification
packages/tasks/src/task/SlackNotifyTask.ts - Slack incoming webhook
packages/tasks/src/task/DiscordNotifyTask.ts - Discord webhook

Files to modify:

packages/tasks/src/common.ts - Export new tasks
packages/tasks/src/task/index.ts - Export from task dir
packages/task-graph/src/task-graph/Workflow.ts - Add .webhookNotify(), .slackNotify(), .discordNotify() methods

Pattern: Follows existing FetchUrlTask pattern. Each task extends Task<Input, Output> with well-defined schemas. They use fetch() internally (available in Bun/Node/Browser).

sroussey:
Adds three side-effecting notification tasks to `@workglow/tasks`, following the `FetchUrlTask` pattern.

Closes #241

## What was added

| File | Purpose |
| --- | --- |
| `packages/tasks/src/task/WebhookNotifyTask.ts` | Generic JSON HTTP POST to any webhook endpoint |
| `packages/tasks/src/task/SlackNotifyTask.ts` | Slack incoming webhook |
| `packages/tasks/src/task/DiscordNotifyTask.ts` | Discord webhook |
| `packages/tasks/src/util/WebhookPost.ts` | Shared internal POST helper (new) |
| `packages/test/src/test/task/NotifyTask.test.ts` | 22 tests (new) |
| `packages/tasks/src/common.ts` | Export + import + `registerCommonTasks()` registration |
| `packages/tasks/README.md` | Three new sections in the existing `### FetchUrlTask` format |

All three run **inline** — no job-queue machinery — and share one POST helper covering `Content-Type: application/json` + `JSON.stringify`, typed error mapping, `Retry-After` parsing, abort/timeout signals, and the private-network entitlement plumbing that `FetchUrlJob.execute` uses. They go through `safeFetch` (never `globalThis.fetch`), so the SSRF boundary and the DNS-rebinding protection are unchanged, and private/loopback destinations still require a scoped `network:private` entitlement re-enforced on every redirect hop.

Each declares `cachePolicy = { kind: "none" }` — the base default of `{ kind: "deterministic" }` would be wrong for a side-effecting send. Category is `"Utility"` (an existing category, matching the other side-effecting sink `DebugLogTask`) rather than inventing a `"Notification"` group.

Discord specifics: a successful post answers **204 No Content**, so the success body is never read and `response.json()` is never called; a 429 may carry `{"retry_after": <seconds>}` in a JSON body instead of a `Retry-After` header, so the body is parsed as a secondary retry source. Payloads are built by omitting absent optional keys entirely — no `username: undefined` in the JSON. (An unset array-typed port materializes as `[]`, which both Slack and Discord reject, so empty arrays are dropped too; this is covered by a test.)

## Two places the issue's own file list was wrong

1. **The issue says to add `.webhookNotify()/.slackNotify()/.discordNotify()` to `packages/task-graph/src/task-graph/Workflow.ts`.** That would invert the dependency graph — `task-graph` sits *below* `tasks`, so it cannot reference task types defined above it. **`Workflow.ts` was not touched.** Instead each new task file ends with TypeScript declaration merging, exactly as `FetchUrlTask.ts` already does:

   ```ts
   declare module "@workglow/task-graph" {
     interface Workflow {
       slackNotify: CreateWorkflow<SlackNotifyTaskInput, SlackNotifyTaskOutput, TaskConfig>;
     }
   }
   Workflow.prototype.slackNotify = CreateWorkflow(SlackNotifyTask);
   ```

   A test proves the augmentation works end to end via `new Workflow().slackNotify({...}).run()`.

2. **The issue says to modify `packages/tasks/src/task/index.ts`.** That file does not exist, and `CLAUDE.md` forbids `index`-named modules. It was **not** created. The single wiring point is `packages/tasks/src/common.ts` (export in the alphabetical block, import, and the class added to the `registerCommonTasks()` array). Platform entrypoints (`browser`/`node`/`bun`/`electron.ts`) needed no edits.

## Secret redaction

Unlike `FetchUrlTask` — public URL plus a *separate* bearer token — a Slack/Discord webhook URL **is** the credential: the token lives in the URL path. Leaking it into a log line or a stored task error is equivalent to leaking an API key, and workflow errors are exactly the thing that gets persisted, surfaced in a UI, and pasted into a bug report.

So the URL is treated as a secret throughout:

- it is **absent from the output schema** (asserted by a test), so it can never land in cached/persisted task output;
- every message this code composes is built from `redactWebhookUrl(url)` → `new URL(url).origin`, never the full URL;
- `safeFetch`'s own errors *do* embed the full URL (e.g. `Refusing to fetch private/internal URL ${url}: ...`), so anything thrown out of the transport is passed through `redactWebhookUrlIn` before it escapes, and the error's `.url` field is rewritten to the origin;
- redaction **preserves classification** — a rewritten error keeps its original `code`, so a permanent SSRF deny stays a `PermanentJobError` and never becomes a retryable `NETWORK_ERROR` that burns the retry budget on an attacker-influenced URL. An error that does not contain the secret is re-thrown by identity, not rewrapped (asserted with `expect(error).toBe(thrown)`).

Slack and Discord failure bodies (`invalid_payload`, `no_service`) carry no secrets and are surfaced in the error message to keep it debuggable; the generic `WebhookNotifyTask` does not echo an arbitrary third-party body into its error message, since an unknown endpoint may reflect the request URL back.

Tests assert that a webhook URL containing `SECRETTOKEN` never appears in a thrown error's message or `.url` — across HTTP failures, network failures, and errors originating inside `safeFetch`.

## Verification

Real commands, real output, run on this branch.

**`bun run build`**
```
Cached:    73 cached, 84 total
  Time:    45.61s
```

**`bun scripts/test.ts task vitest`** — the section covering these tasks
```
Running all tests in sections [task] — 73 file(s)
 Test Files  73 passed (73)
      Tests  1154 passed | 24 skipped (1178)
   Duration  267.88s
```
The new `NotifyTask.test.ts` contributes 22 of those. Run alone:
```
 Test Files  1 passed (1)
      Tests  22 passed (22)
```

**`bun run build:types`** — this reported `FULL TURBO` from cache, so it was re-run with `--force` to get a genuine result:
```
$ npx turbo run build-types --force --concurrency=15
Cached:    0 cached, 41 total
  Time:    43.839s
```
That includes `@workglow/test:build-types`, which typechecks the new test file.

**`bun scripts/test.ts graph vitest`** — extra check, not requested, because `registerCommonTasks()` gained three entries and the graph suite calls it:
```
 Test Files  76 passed (76)
      Tests  747 passed (747)
```

`bun run format` (eslint --fix + prettier) is clean; the pre-commit hook re-ran both and produced no further changes.

### Test coverage

Mocking swaps the SafeFetch impl (`registerSafeFetch(mockSafeFetch)` in `beforeAll`, restored in `afterAll`) rather than patching global fetch, copying the setup from `FetchTask.test.ts`.

Happy path per task asserting method `POST`, JSON content type, and the exact serialized payload · header merging · 1KB response truncation · optional fields omitted (including the empty-array case) · Discord 204 success with no body read or JSON parse · 4xx permanent error surfacing the Slack body · 429 with `Retry-After` producing a retryable error with a retry date · the Discord JSON-`retry_after` variant · network rejection → retryable · missing-URL configuration error · secret redaction across three failure modes · permanent-error pass-through by identity · URL absent from all output schemas · `registerCommonTasks()` registration of all three · schema validity · `new Workflow().slackNotify({...}).run()` round trip.

One deviation from the issue-triage brief: it suggested asserting `TaskRegistry.getTask("SlackNotifyTask")`. That API does not exist — the registry exposes a `Map`, so the test uses `TaskRegistry.all.get("SlackNotifyTask")`.

---


sroussey:
Applied the two requested changes and moved this to **draft**.

**Changes** (`0a33c7c1`):
- `category` on all three notify tasks → `"Notification"` (was `"Utility"`/`"Output"`)
- port renamed `credential_key` → `url_credential_key` on `WebhookNotifyTask`, `SlackNotifyTask`, `DiscordNotifyTask` and the shared `WebhookPost` helper. `format: "credential"` and `"x-ui-hidden": true` are unchanged. `FetchUrlTask`'s own `credential_key` is untouched — it is a different port on a different task.

**Why draft, not merge:** open architectural question about *where* these belong. Naming specific notification targets (Slack, Discord) inside core `packages/tasks/` may be the wrong layer — the same argument that puts concrete AI vendors in `providers/*` rather than in `@workglow/ai` applies here. The generic `WebhookNotifyTask` is plausibly core; `SlackNotifyTask` / `DiscordNotifyTask` are vendor-shaped and arguably want their own package(s) under `providers/`, so a consumer that never notifies Slack doesn't carry its payload shaping.

Counter-argument for keeping them in `tasks`: they have zero third-party dependencies — all three are `fetch` against a URL with a JSON body, so there is no dependency-weight reason to split, and `providers/*` is currently structured around AI model providers with `./ai` + `./ai-runtime` sub-paths, which is not the shape a notification target wants.

Not blocking on this — leaving as draft until the layering call is made.

Verification on the current head: `bun run build` 84/84 (plus a forced uncached `build-types` 37/37), `bun scripts/test.ts task vitest` 1158 passed / 24 skipped, `NotifyTask.test.ts` 26 passed, prettier clean.

---

Issue #568, opened 2026-06-12
Decompose Task.ts, BaseTabularStorage.ts, and TaskRunner.ts


sroussey:
Roadmap: `prd/roadmap/libs.md`

Three oversized flagship files:
- `Task.ts` (1148 LOC) → TaskCloneOps + TaskJsonOps + TaskCacheOps
- `BaseTabularStorage.ts` (1058 LOC) → CursorCodec + QueryHelpers + SubscriptionMixin
- `TaskRunner.ts` (893 LOC) → RunLifecycle

Mechanical decomposition along functional seams.

Grades: `prd/analysis/grades/libs.md` recommendation #2

sroussey:
2026-08-24 review — measurements at `de0d4f2df`, and an argument for doing this **now**.

| file | LOC | state |
|---|---|---|
| `TaskRunner.ts` | **1624** | unchanged this window |
| `Task.ts` | 1042 | done, holding |
| `BaseTabularStorage.ts` | **696** | done, holding three cycles |

`find . -name 'RunLifecycle*'` returns nothing.

`TaskRunner.ts` is flat this cycle only because `packages/task-graph/src` had a **zero-byte source diff** for the entire window — not because anything was extracted. It was ~934 when this issue was filed and grew in every prior cycle.

That makes this the lowest-conflict moment the extraction will ever have: no in-flight work in the package to rebase against. The precedent already worked and held — `a7d72dd1` took `BaseTabularStorage.ts` 1142 → 696 with signature-preserving one-line delegations, and it is still exactly 696 three cycles later.

Ask: extract the run-lifecycle block into `RunLifecycle.ts` with the same shape, keeping every public signature. Since two of the three named files are done, retitling to what remains would make the issue readable again.

Separately, `FetchUrlTask.ts` grew **1507 → 1640** this window and nothing tracks it.

---
_Generated by [Claude Code](https://claude.ai/code)_

sroussey:
**Re-verified 2026-08-31 against `origin/main` @ `9a80450`. Growth resumed on the one file this issue is named for.**

| file | 2026-08-17 | 2026-08-24 | **2026-08-31** |
|---|---|---|---|
| `packages/task-graph/src/task/TaskRunner.ts` | 1624 | 1624 | **1695** (+71) |
| `packages/task-graph/src/task/Task.ts` | 1042 | 1042 | 1042 |
[...]

PR #682, opened 2026-08-05
refactor: decompose BaseTabularStorage.ts and Task.ts along functional seams

sroussey:
Refs #568

Mechanical decomposition of two of the three flagship files named in the issue. **`TaskRunner.ts` is deliberately deferred** — see below.

## Why TaskRunner.ts is deferred

Open PR #621 (`full-stream`, 108 files, 83 commits) rewrites `TaskRunner.ts` with **+243/−8 across 7 hunks**, including an insertion **inside `handleStart`'s cache-resolution block** — squarely inside the `RunLifecycle` cluster the issue proposes to extract.

Moving `handleStart` into a new file now would create a delete/modify conflict on brand-new logic, and would force whoever rebases #621 to hand-port those hunks into a file that did not exist when they were written. That is a bad trade for a purely mechanical refactor.

**Recommendation:** re-plan the `TaskRunner.ts` split against the *post-#621* file. The seam will look different once streaming lands, and the split is cheap to redo against the settled version.

By contrast:
- `BaseTabularStorage.ts` is untouched by all three open PRs.
- `Task.ts` gets only +45/−4 from #621, confined to the schema-validation region (roughly lines 837-967). Every hunk in this PR is textually disjoint from it — the closest edits are at old-file lines 568-634 and 976-1007. The `schemaWithoutPorts` / `validateInput` region is not touched.

## The issue's "SubscriptionMixin" does not map onto real code

The issue proposes splitting `BaseTabularStorage.ts` into `CursorCodec + QueryHelpers + SubscriptionMixin`. The first two map cleanly; the third does not.

Subscriptions in this file are ~45 LOC total: a ~10-line `subscribeToChanges` that just **throws** ("override in subclasses"), plus thin `on`/`off`/`once`/`emit`/`waitOn` delegation to an `EventEmitter`. There is no machinery to extract — the real subscription logic lives in the concrete backends and in `packages/storage/src/util/{Hybrid,Polling}SubscriptionManager.ts`, which are already separate modules. Extracting a "mixin" here would move a throw statement into its own file and add an indirection for nothing.

**What was split instead** — four clusters that are genuinely large and genuinely self-contained:

| New module | What moved | LOC |
|---|---|---|
| `cursorValues.ts` | `toCursorValue`, `compareKeyValues` (pure, module-level) | 69 |
| `keysetPage.ts` | the `runPage` engine as `runKeysetPage(deps, criteria, request)`, plus `buildEffectiveOrderBy` / `sortInMemory` / `applyKeysetFilter` / `buildCursor` bodies | 321 |
| `tabularValidation.ts` | the five `validate*` bodies, taking schema properties as their first argument | 154 |
| `tabularSchemaSetup.ts` | constructor helpers: schema splitting, column-name validation, index normalization + unique-index dedup, auto-generated-key detection, `determineGenerationStrategy` | 245 |

`cursorValues.ts` + `keysetPage.ts` together are the issue's "CursorCodec"; `tabularValidation.ts` is its "QueryHelpers"; `tabularSchemaSetup.ts` is an additional seam the issue did not name but which is where the constructor's bulk actually lives.

## LOC, before and after

The issue's numbers are stale — the files have grown since it was filed.

| File | Issue says | Actual before | After |
|---|---|---|---|
| `BaseTabularStorage.ts` | 1058 | **1142** | **693** (−449, −39%) |
| `Task.ts` | 1148 | **1169** | **1003** (−166, −14%) |
| `TaskRunner.ts` | 893 | **987** | *(deferred, unchanged)* |

New modules: storage 69 + 321 + 154 + 245; task-graph 119 (`TaskCloneOps`) + 81 (`TaskCacheOps`) + 83 (`TaskJsonOps`).

**Follow-up:** `Task.ts` is still ~1000 LOC. The clone/JSON/cache seam the issue names is now fully harvested; getting it meaningfully below that needs an **unlisted** input-ops seam (`setInput` / `addInput` / defaults handling, ~240 LOC), which is a separate judgement call about whether input merging belongs outside the class. I did not over-cut to hit a number.

## Correctness constraints held

These are what make the refactor safe for out-of-repo subclasses, and none of them are covered by tests in this repo:

- **Every extracted `protected` method keeps its signature on the class** as a one-line delegation. Bodies moved to free functions; the methods stayed. A scripted comparison of the class member surface before vs after reports **69 → 69 members for `BaseTabularStorage` and 86 → 86 for `Task`, with zero removed, changed, or added**. Same for `Task`'s `getCachePolicy` (subclass-overridden), `getCacheVersion`, `hasDeterministicId`, and the `cacheable` getter.
- **Virtual dispatch preserved.** `runKeysetPage` takes a `KeysetPageDeps<Entity>` bag of closures bound to the instance (the `StreamProcessor.ts` pattern) and never calls a sibling free function directly. Subclasses override `query` / `getAll` / `generateKeyValue` / `sortInMemory`; had the engine short-circuited to a free function, an out-of-repo override would have silently stopped taking effect with no test failure here.
- **`Task.toJSON` remains a method** — `GraphAsTask`, `FallbackTask` and `ArrayTask` override it and call `super.toJSON()`. Its body became `buildTaskJson(task)`, which reads statics via `task.constructor as typeof Task` and calls `canSerializeConfig()` on the instance so subclass overrides still gate serialization. `Task` is imported as a **type only**, so no value cycle is introduced.
- **The literal NUL byte (U+0000) in the unique-index tuple key is preserved.** It moved into `tabularSchemaSetup.ts`; the byte was verified in the committed git blob (2 NUL bytes, in `join("\0")` on both lines), not just on disk — including after the pre-commit prettier/eslint hooks ran.
- **Public API unchanged.** `TABULAR_REPOSITORY`, `ClientProvidedKeysOption` and `KeyGenerationStrategy` stay physically in `BaseTabularStorage.ts`, which `common.ts` re-exports wholesale. The new helper modules are internal and were **not** added to any `common.ts` / `index.ts`.
- Zero subclass edits; zero drive-by cleanups (the existing `as unknown as Array<keyof Entity>` casts were left alone).

## Verification

All commands run from a clean build of this branch.

| Command | Result |
|---|---|
| `bun scripts/test.ts storage vitest` | **61 files passed, 1830 passed / 23 skipped** |
| `bun scripts/test.ts queue vitest` | **22 files passed, 505 passed / 26 skipped** |
| `bun scripts/test.ts rag vitest` | **24 files passed, 225 passed** |
| `bun scripts/test.ts task graph vitest` | **148 files passed, 1879 passed / 24 skipped** |
| `bun run build:types` | **41/41 tasks successful** |
| `bun run build` | **84/84 tasks successful** |
| `npx eslint` (9 touched files) | clean, no output |

### Confirming the non-SQL backends actually exercised the new keyset engine

The default `runPage` path is the riskiest part of this change, and it is exercised **only** by the non-SQL backends — SQLite / Postgres / DuckDB / Supabase all override `getPage`, and their integration suites skip without credentials. A green aggregate run therefore proves nothing on its own.

To confirm, the relevant files were re-run with `--reporter=verbose` and the full log inspected per test (not just the summary):

```
npx vitest run --reporter=verbose \
  packages/test/src/test/storage-tabular/InMemoryTabularStorage.test.ts \
  packages/test/src/test/storage-tabular/FsFolderTabularStorage.integration.test.ts \
  packages/test/src/test/storage-tabular/IndexedDbTabularStorage.integration.test.ts \
  packages/test/src/test/storage-tabular/SharedInMemoryTabularStorage.test.ts

Test Files  4 passed (4)
     Tests  395 passed | 14 skipped (409)
```

Filtering that log for cursor/paging test names gives **66 individually-passing (`✓`) keyset tests**, distributed across the three backends that inherit the default path:

- `InMemoryTabularStorage` — 27
- `IndexedDbTabularStorage` — 29
- `FsFolderTabularStorage` — 10

Named cases include *"iterates the entire table by following nextCursor"*, *"paginates filtered results via queryPage"*, *"paginates by a non-PK orderBy with primary key as tiebreaker"*, *"paginates compound (nullable DESC, non-null ASC) — mixed direction with NULLs trailing"*, and *"is stable under concurrent inserts that sort before the cursor"*.

All 14 skipped tests were enumerated and **none are pagination tests** — they are `withTransactionRollback` (backends without native transactions), `vectorColumnFormat`, and FsFolder's `inListCriterion` / `countMatchesQuery` group. So no keyset test was silently skipped.

The FsFolder tests are also what cover the `StorageUnsupportedError` → `getAll` fallback branch inside `runKeysetPage`, since FsFolder does not implement `query()`.

---


sroussey:
## The `TaskRunner.ts` deferral is confirmed correct

Maintainer confirms **#621 (`full-stream`) is alive and mid-rebase**. That settles the open question on this PR: the `TaskRunner.ts` split stays deferred, and #568 stays open on `Refs` rather than closing.

Verified the deferral actually buys what it claims — trial-merged this branch and then #621 on top, against `origin/main` at `0c90d19b`:

```
#621 → main + #682 :  0 conflicts
#621 → main        :  0 conflicts   (baseline)
```

So merging this PR leaves #621's rebase exactly as clean as it is today. The `Task.ts` hunks here are textually disjoint from #621's schema-validation edits, as the body claimed, and `BaseTabularStorage.ts` is untouched by #621 — now measured, not assumed.

Had `TaskRunner.ts` been split here, this would have been a delete/modify conflict on `handleStart`'s cache-resolution block against brand-new #621 logic — the exact cost the deferral avoided.

Ready to merge, and free to go ahead of #683 (which is the one PR in the batch that *does* conflict with #621 — 5 files; see the note there).

---

PR #684, opened 2026-08-05
refactor(job-queue): collapse per-backend queue adapters onto wrapQueueStorage

sroussey:
## What

Deletes the five hand-written `MessageQueue` + `JobStore` adapter pairs (InMemory, IndexedDb, Postgres, Sqlite, Supabase — ten files, ~1,400 LOC) and makes every `createXQueue` factory return `wrapQueueStorage(new XQueueStorage(...))`, so the `IMessageQueue`/`IJobStore` facade semantics exist in exactly one place.

**Net: +163 / −1,664 across 31 files.**

## Why

The per-backend classes were byte-identical copies of the generic wrapper (modulo names) that had silently drifted behind its fixes. Deleting them structurally removes four latent bugs the copies still carried:

- `claim.ack()` / `claim.fail()` fell back to `current.output` / `current.error` — resurrecting the **prior attempt's** result on a retried job (the wrapper was fixed; the copies never were).
- `claim.retry()` didn't clear `abort_requested_at`, so an abort requested during a failed attempt survived into the next one.
- `sendBatch()` lacked the single-fingerprint guard (every body would dedup against the first row).
- The IndexedDb fingerprint scan was unbounded, where the wrapper bounds it and warns on exhaustion.

## How backend behavior is preserved

- `WrappedJobStore` now duck-probes the storage for native `saveStatus` / `getMany` / `completeWithResult` / `failWithError` fast paths (single atomic statements on the SQL backends), the same way it already probed `findActiveByFingerprint` — so the SQL backends keep their one-round-trip writes and indexed lookups.
- `findActiveByFingerprint` enforces queue-name scoping when the storage exposes its `queueName`, matching the native SQL `WHERE queue = ?` lookups. Previously only the InMemory adapter enforced this on the non-SQL path; the shared `JobStoreExtensions` suite pins it for all backends.
- `WrappedJobStore` exposes `backingStorageName` so the SQS/Cloudflare adapters keep their "non-durable in-memory store" pairing warning, which previously keyed off the deleted class's constructor name.

## API notes

- The ten concrete class names were exported from each package's barrel but had **no consumers** outside their own factories and tests — verified by grep across libs, sec, embarc-data, embarc, and examples. They are removed without aliases.
- `createXQueue` return types narrow from the concrete classes to `IMessageQueue<JobStorageFormat<Input, Output>>` / `IJobStore<Input, Output>`.

## Testing

- `bun scripts/test.ts vitest unit queue cloudflare` — 18 files, 221 passed
- `bun scripts/test.ts vitest integration queue cloudflare` — 10 files, 307 passed (Postgres/Supabase live suites env-gated as usual)
- `bun scripts/test.ts vitest unit graph task` — 139 files, 1,843 passed (downstream consumers)
- `bun run build:packages` clean; eslint/prettier clean

https://claude.ai/code/session_01TQFDxK7AKh4w4DCjRJUpQm

---
_Generated by [Claude Code](https://claude.ai/code/session_01TQFDxK7AKh4w4DCjRJUpQm)_


sroussey:
**Code review (xhigh, --fix) — 10 finder angles + 5 adversarial verifiers. Fixes pushed in 9c6321a8.**

**Applied (converged findings):**
- Six independent angles flagged the duck-typed probes (`QueueStorageFastPaths` shadow interface, `backingStorageName === "InMemoryQueueStorage"`, structural `queueName` cast) as decorator-opaque: wrapping any backend in `TelemetryQueueStorage` silently downgraded `failWithError`/`completeWithResult` to read-then-write compositions (re-opening the lost-update window the native statements exist to close) and un-batched `getMany`. The four fast paths plus `queueName`/`durable` are now declared optional members on `IQueueStorage` (the `findActiveByFingerprint` precedent), the wrapper probes typed members, and Telemetry mirrors them conditionally.
- The in-memory pairing warning now keys on a declared `durable` flag instead of a constructor-name string (survives decorators and minified bundles), via a shared `warnIfNonDurableJobStore`; the duplicated defer-backoff/clamp/fallback helpers in SQS/Cloudflare moved to one `messageQueueAdapterSupport` module.
- `SqliteQueueStorage.saveProgress(null)` stored the JSON text `'null'` where the deleted `SqliteJobStore` coerced to `{}`; it now stores SQL NULL (matching its own `add`/`finalize` and the other backends).
- `JobQueueServer` startup diagnostics printed `"WrappedMessageQueue"` for every collapsed backend; the wrapper now exposes `backingStorageName` on the message queue and the log names the real backend.
- Named `QueuePair` return type for the five factories; stale comments corrected (SQS partial-failure comment claimed successful rows keep `visible_at = now`, which stops being true once #686's `add()` preserves caller-set visibility; `markEnqueueDeferredMany` "backends can override" wording; `IJobStore.create` JSDoc); test describes renamed off the deleted class names.

**Investigated and cleared (no change):**
- *`retry()` clearing `abort_requested_at` loses user aborts* — refuted. Every backend's storage layer already hard-nulls the column on the retry branch (`abort_requested_at = NULL` in the SQL, with the same rationale comment), `next()` clears it again at re-claim, and abort-caused failures route to `failJob` (AbortSignalJobError is permanent), never `retry()`. The deleted claims' `...current` spreads were dead code for this column; semantics are per-lease by design and contract-tested at main.
- *`sendBatch` fingerprint `RangeError`* and *`ack()` no longer falling back to `current.output`* — intentional, documented in the wrapper, and test-pinned; the old fallbacks resurrected prior-attempt values.
- *queueName guard blocks cross-queue dedup* — cross-queue lookup was never the `IJobStore` contract ("within this queue"), InMemory already had the identical guard, and a cross-queue hit would itself have been a bug (dedup id returned without enqueueing).

Verified after the changes: `build:packages` clean, `tsc -b` clean, queue/aws/cloudflare vitest sections 554 passed.

---

PR #685, opened 2026-08-05
feat(providers): fill the Usage seam per provider (#624 Phase 2)

### LINKED ISSUE #624: Add uniform usage/token telemetry across all AI providers
A uniform, provider-agnostic usage/token telemetry channel, so cost is computable from a workflow run.

Delivered in two phases. **Phase 1 has landed; Phase 2 is what remains.**

## Phase 1 — the generic seam ✅ (merged in #680)

The four design questions from the original issue are settled:

1. **Shape** — `Usage` in `packages/task-graph/src/task/StreamTypes.ts`: `input`, `output`, `cached`, `cacheWrite`, `reasoning`, `total`, plus an `extra` bag for provider-specific counters. Every field is `number | undefined`, where **`undefined` means "not reported", never "zero"** — collapsing the two understates spend. Re-exported from `@workglow/ai`.
2. **Transport** — an optional `usage` on `StreamFinish`, as a **sibling of `data`, never inside it**, so the `{}`-for-delta-streams, full-Output-for-one-shot, and `data.object`-for-json-mode conventions all stay intact. A new stream-event arm was rejected (every `switch` over `StreamEvent` would need a case, and ordering against `finish` would be ambiguous); the run result was rejected because `AiProviderRunFn` returns `Promise<void>` by contract. Survives the worker boundary via structured clone with no protocol change.
3. **Surfacing** — a reserved output key mirroring the existing `refusal` precedent: folded by `StreamEventAccumulator.applyUsage` and `StreamProcessor`, undeclared on any `outputSchema()`, kept off dataflow edges. One difference from refusal, deliberate: refusal **blocks** a cache save, usage is **stripped** and the value saved — a cache hit costs zero tokens, so replaying counts would invent spend. Telemetry is recorded in one place (`AiTask.execute` / `StreamingAiTask.executeStream`) via `UsageTelemetry.ts`, using OTel gen-ai semconv attributes; `ITelemetryProvider` unchanged.
4. **Multi-turn** — aggregated consumer-side so providers stay stateless: `AiChatWithKbTask` merges each agent-loop turn, `StructuredGenerationTask` sums across validation retries (a rejected attempt is still a billed request).

## Phase 2 — per-provider fill-in ⬅ remaining work

Every provider currently emits `usage: undefined`, so nothing reports real numbers yet. Each run-fn maps its own finish payload into `Usage` and attaches it to the `finish` it already emits.

**Blocked on #641** (`ai-provider-cache-checkpoints`), which edits ~15 of the same run-fns. #641 has now been rebased onto main; Phase 2 stacks on top of it.

**Usage already flows past the loop and is discarded — no request change needed**

- [ ] `@workglow/anthropic` — `usage.input_tokens` / `output_tokens` / `cache_read_input_tokens` → `cached` / `cache_creation_input_tokens` → `cacheWrite`
- [ ] `@workglow/openai` (Responses) — `usage.input_tokens` / `output_tokens` / `input_tokens_details.cached_tokens` / `cache_write_tokens` / `output_tokens_details.reasoning_tokens`
- [ ] `@workglow/google-gemini` — `usageMetadata.promptTokenCount` / `candidatesTokenCount` / `cachedContentTokenCount` / `thoughtsTokenCount`
- [ ] `@workglow/ollama` — the `done: true` chunk's prompt/eval counts

**Needs `stream_options: { include_usage: true }` on the request**

- [ ] `@workglow/xai`
- [ ] `@workglow/deepseek` — also `prompt_cache_hit_tokens` → `cached`
- [ ] `@workglow/openrouter` — also `cost` → `extra`
- [ ] `@workglow/llamacpp-server`

**No usage available — leave unreported (`undefined`, not `0`)**

- [ ] `@workglow/node-llama-cpp`
- [ ] `@workglow/huggingface-transformers`
- [ ] `@workglow/chrome-ai`
- [ ] `@workglow/tf-mediapipe`

## Known gap, not yet addressed

`StructuredOutputValidationError` drops the accumulated usage when *every* validation attempt fails — the most expensive outcome reports nothing. Fixing it requires an error-shape change.

sroussey:
Closes #624

## Stacked PR — merge after #641

This branches from **`ai-provider-cache-checkpoints`** (PR #641), not `main`, and targets that branch. **It must merge after #641.**

Phase 1 (#680, already on `main`) added the generic seam — `Usage`, `USAGE_OUTPUT_KEY`, `mergeUsage`, and `usage` as an optional **sibling of `data`** on `StreamFinish` — with every provider emitting `usage: undefined`. This PR is the per-provider fill-in: each run-fn maps its own terminal frame into `Usage` and attaches it to the `finish` event it already emits. The consumer side (`StreamEventAccumulator.applyUsage` / `StreamProcessor`) is untouched.

`usage` is always attached as a sibling of `data`, never inside it — folding token counts into `data` would break the `{}`-for-delta-streams, full-Output-for-one-shot, and `data.object`-for-json-mode conventions all at once.

## The `undefined` vs `0` rule

**`undefined` means "the provider did not report this figure". It never means zero.**

A model that billed 0 cached tokens and a model that says nothing about caching are different facts, and collapsing them to `0` silently understates spend — the reader of a cost report cannot tell "this request used no cache" from "we have no idea whether it did". So an absent, `null`, `NaN`, or non-numeric wire field maps to `undefined`, while a genuinely reported `0` survives as `0`. Providers with no usage at all leave it absent entirely.

Usage is also always read from the provider's **own terminal frame** — never derived by counting our own emitted output, which would drift from what the provider actually bills.

## Per-provider mapping

| Provider | Source frame | input | output | cached | cacheWrite | reasoning | total | extra |
|---|---|---|---|---|---|---|---|---|
| **anthropic** | `message_start` + `message_delta` | `input_tokens` | `output_tokens` | `cache_read_input_tokens` | `cache_creation_input_tokens` | `output_tokens_details.thinking_tokens` | — | — |
| **openai** (Responses) | `response.completed` | `input_tokens` | `output_tokens` | `input_tokens_details.cached_tokens` | `input_tokens_details.cache_write_tokens` | `output_tokens_details.reasoning_tokens` | `total_tokens` | — |
| **openai** (embeddings) | non-streaming result | `usage.prompt_tokens` | — | — | — | — | `usage.total_tokens` | — |
| **google-gemini** | last non-null `chunk.usageMetadata` | `promptTokenCount` | `candidatesTokenCount` | `cachedContentTokenCount` | — | `thoughtsTokenCount` | `totalTokenCount` | — |
| **ollama** | `done: true` chunk | `prompt_eval_count` | `eval_count` | — | — | — | — | — |
| **xai** | `include_usage` frame | `prompt_tokens` | `completion_tokens` | `prompt_tokens_details.cached_tokens` | — | `completion_tokens_details.reasoning_tokens` | `total_tokens` | — |
| **deepseek** | `include_usage` frame | `prompt_tokens` | `completion_tokens` | `prompt_cache_hit_tokens` | — | `completion_tokens_details.reasoning_tokens` | `total_tokens` | `promptCacheMissTokens` |
| **openrouter** | `include_usage` frame | `prompt_tokens` | `completion_tokens` | `prompt_tokens_details.cached_tokens` | — | `completion_tokens_details.reasoning_tokens` | `total_tokens` | `cost` |
| **llamacpp-server** | `include_usage` frame | `prompt_tokens` | `completion_tokens` | `prompt_tokens_details.cached_tokens` | — | `completion_tokens_details.reasoning_tokens` | `total_tokens` | — |
| **huggingface-inference** | terminal chunk, if volunteered | `prompt_tokens` | `completion_tokens` | `prompt_tokens_details.cached_tokens` | — | `completion_tokens_details.reasoning_tokens` | `total_tokens` | — |

### Providers that report nothing — `usage` stays absent

| Provider | Why |
|---|---|
| **node-llama-cpp** | Runs the model locally; its prompt API surfaces no token accounting. |
| **huggingface-transformers** | transformers.js runs in-process and reports no counts. |
| **chrome-ai** | The Chrome built-in AI API exposes only quota measures (`inputUsage`), not billing counters. |
| **tf-mediapipe** | Vision pipelines; no token concept. |

These are left absent with a short comment recording why, so the next reader does not "fix" it by emitting zeros or by counting our own emitted tokens.

### Notes on specific decisions

- **Anthropic reports no `total`.** Synthesizing one from the parts would misreport cache-discounted input as if it had been billed in full, and `Usage.total` is documented as never synthesized. Both frames restate **cumulative** figures, so the collector keeps the newest value per field rather than summing; a `null` on `message_delta` means "not restated here" and leaves the earlier value intact rather than erasing it.
- **DeepSeek's `prompt_cache_miss_tokens`** has no normalized slot but is what makes its cache-miss-priced cost reconstructable, so it rides in `extra`.
- **OpenRouter's `cost`** is a price, not a token count. It is worth carrying because OpenRouter routes to whichever upstream is cheapest, which the caller cannot reconstruct locally.
- **Ollama field names were verified** against the installed SDK's `ChatResponse` / `GenerateResponse` types rather than taken on trust.

## Shared plumbing

`packages/ai/src/provider-utils/UsageMapping.ts` normalizes the two OpenAI wire shapes and exposes the primitives provider packages compose their quirks from (`toUsageCount`, `usageOrUndefined`, `usageExtra`, `OPENAI_STREAM_USAGE_OPTIONS`).

Both shared accumulators now **return** the usage they read instead of discarding it:

- `accumulateOpenAIResponsesStream` handles `response.completed` (plus the `incomplete` / `failed` terminals, which still consumed tokens) rather than dropping them into the `default:` arm. One change covers OpenAI TextGeneration, ToolCalling, Rewriter and Summary.
- `accumulateOpenAIChatStream` reads `chunk.usage` **before** skipping a choice-less chunk, and takes an optional mapper so DeepSeek/OpenRouter can supply their own.

### `include_usage` safety

xAI, DeepSeek, OpenRouter and llama.cpp-server needed `stream_options: { include_usage: true }` added to the request. That opt-in appends a final chunk whose **`choices` array is empty** — exactly the shape the `*_StreamChunkSafety` suites exist to guard. Every delta read already optional-chains `chunk.choices?.[0]`, and usage is read before the choice-less chunk is skipped. Those suites were extended rather than duplicated.

llama.cpp-server parses SSE itself, so `IChatCompletionDelta` gained a `usage` field and the parser now admits a usage-only frame on its own merit — it previously required a content delta, tool calls, or a finish reason, which silently dropped the usage chunk.

## Tests

A shared contract assertion at `packages/test/src/contract/ai-provider/assertions/usageNormalization.ts` (sibling of `sessionReuse.ts`) holds every provider's mapper to the same rules: no payload maps to `undefined`, unreported counters come back `undefined` rather than `0`, a genuinely reported `0` survives, and only numbers and `undefined` reach the normalized shape. `ProviderUsageNormalization.test.ts` runs all seven mappers through it, so the zero-vs-absent rule is enforced in one place instead of being restated per provider.

Per-provider suites were extended in place: `Xai_StreamChunkSafety`, `OpenRouter_StreamChunkSafety`, `LlamaCppServer_TextGenerationStream`, `Ollama_TextGenerationStream`, `DeepSeek_ToolCalling`, `Anthropic_SamplingParams`, and `OpenAIShapedResponses`.

## Verification

All commands run against the committed branch state.

```
$ bun run build
Cached:    84 cached, 84 total

$ bun run build:types
Cached:    41 cached, 41 total

$ bun scripts/test.ts ai provider provider-nodellama vitest unit
 Test Files  1 failed | 63 passed (64)
      Tests  1 failed | 669 passed | 1 skipped (671)

$ bun scripts/test.ts provider-api vitest unit
 Test Files  38 passed (38)
      Tests  431 passed (431)
```

`provider-api` went from 413 to 431 tests; the `ai`/`provider` sweep gained the shared-contract and Responses-usage suites.

### Inherited failure — not from this PR

The single failure is `packages/test/src/test/ai-provider/OpenAIGeminiCheckpointParams.test.ts` → "evicts and retries inline once on a NOT_FOUND".

It **already fails on the base branch**: commit `5d9088f3 fix(gemini): tighten reactive CachedContent NOT_FOUND matcher` (the rebased `3206e7de`) tightened the CachedContent NOT_FOUND matcher without updating that older test. `5d9088f3` is the second commit of `ai-provider-cache-checkpoints`, and this PR's diff touches neither `Gemini_CachedContentFallback.ts` nor that test file — the stack trace runs entirely through `generateGeminiStreamWithCacheFallback`, which is untouched here. It is reported on #641 and awaiting the owner's decision; it is deliberately not fixed in this PR.

Only the `unit` test kind was run. Integration suites were not run locally, since the preloader unlocks real API keys and `test-vitest-ai-provider-api` makes real billable `gpt-image-2` calls.

---

### DIFF
--- a/packages/ai/src/provider-utils.ts
+++ b/packages/ai/src/provider-utils.ts
@@ -25,6 +25,7 @@ export * from "./provider-utils/CloudProviderClient";
 export * from "./provider-utils/OpenAIShapedChat";
 export * from "./provider-utils/OpenAIShapedResponses";
 export * from "./provider-utils/OpenAIShapedJsonSchema";
+export * from "./provider-utils/UsageMapping";
 export * from "./provider-utils/RefusalHelpers";
 export * from "./provider-utils/IBackendsTransport";
 export * from "./provider-utils/localUrl";
--- a/packages/ai/src/provider-utils/OpenAIShapedChat.ts
+++ b/packages/ai/src/provider-utils/OpenAIShapedChat.ts
@@ -4,11 +4,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { StreamEvent } from "@workglow/task-graph";
+import type { StreamEvent, Usage } from "@workglow/task-graph";
 import { parsePartialJson } from "@workglow/util/worker";
 import type { ToolCallingTaskOutput } from "../task/ToolCallingTask";
 import type { ToolCalls, ToolDefinition } from "../task/ToolCallingUtils";
 import { buildToolDescription, sanitizeToolArgs } from "../task/ToolCallingUtils";
+import { mapOpenAIChatUsage } from "./UsageMapping";
 
 /**
  * Shared helpers for providers that expose an OpenAI-compatible chat-completions
@@ -135,14 +136,28 @@ interface ToolCallAccumulatorEntry {
  * `{ text: "", toolCalls: [] }` so the final output satisfies
  * {@link ToolCallingTaskOutput} even when the model streams only
  * `tool_calls` (no `content` deltas).
+ *
+ * Returns the stream's token accounting when the request opted in with
+ * `stream_options: { include_usage: true }`, else `undefined`. That usage
+ * arrives on a final chunk with an **empty** `choices` array, so it is read
+ * before the choice-less chunk is skipped. The caller attaches it to its
+ * `finish` event as a sibling of `data`.
+ *
+ * `mapUsage` defaults to the standard OpenAI-shape mapping; a provider that
+ * reports counters beyond that common set (DeepSeek's cache hit/miss split,
+ * OpenRouter's `cost`) passes its own mapper instead.
  */
 export async function accumulateOpenAIChatStream(
   stream: AsyncIterable<any>,
-  emit: (event: StreamEvent<ToolCallingTaskOutput>) => void
-): Promise<void> {
+  emit: (event: StreamEvent<ToolCallingTaskOutput>) => void,
+  mapUsage: (raw: unknown) => Usage | undefined = mapOpenAIChatUsage
+): Promise<Usage | undefined> {
   const toolCallAccumulator = new Map<number, ToolCallAccumulatorEntry>();
+  let usage: Usage | undefined;
 
   for await (const chunk of stream) {
+    usage = mapUsage(chunk?.usage) ?? usage;
+
     const choice = chunk?.choices?.[0];
     if (!choice) continue;
 
@@ -188,4 +203,6 @@ export async function accumulateOpenAIChatStream(
       });
     }
   }
+
+  return usage;
 }
--- a/packages/ai/src/provider-utils/OpenAIShapedResponses.ts
+++ b/packages/ai/src/provider-utils/OpenAIShapedResponses.ts
@@ -4,10 +4,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { StreamEvent } from "@workglow/task-graph";
+import type { StreamEvent, Usage } from "@workglow/task-graph";
 import type { ToolCalls, ToolDefinition } from "../task/ToolCallingUtils";
 import { buildToolDescription, sanitizeToolArgs } from "../task/ToolCallingUtils";
 import { parseToolArgs } from "./OpenAIShapedChat";
+import { mapOpenAIResponsesUsage } from "./UsageMapping";
 
 /**
  * Shared helpers for the OpenAI **Responses** API (`client.responses.create`),
@@ -233,17 +234,23 @@ interface ResponsesToolCallEntry {
  * own `finish` event after this returns (per the streaming convention). Generic
  * over the output type so non-tool-calling run-fns (TextGeneration / Rewriter /
  * Summary) can pass their own `emit` under `strictFunctionTypes`.
+ *
+ * Returns the token accounting off the terminal `response.completed` event —
+ * the Responses API reports usage there unprompted — or `undefined` when the
+ * stream ended without one. The caller attaches it to its `finish` event as a
+ * sibling of `data`.
  */
 export async function accumulateOpenAIResponsesStream<Output = Record<string, any>>(
   stream: AsyncIterable<any>,
   emit: (event: StreamEvent<Output>) => void
-): Promise<void> {
+): Promise<Usage | undefined> {
   // Keyed by the output item's stable id (`item.id` on `output_item.{added,done}`,
   // `item_id` on `function_call_arguments.delta`). Falls back to `output_index`
   // only for gateways that omit both — keying purely on `output_index` collapses
   // concurrent function calls that share (or lack) the index onto one slot,
   // silently dropping every call but the last.
   const toolCalls = new Map<string, ResponsesToolCallEntry>();
+  let usage: Usage | undefined;
 
   const emitToolCall = (entry: ResponsesToolCallEntry): void => {
     const parsed = parseToolArgs(entry.arguments);
@@ -341,10 +348,22 @@ export async function accumulateOpenAIResponsesStream<Output = Record<string, an
         break;
       }
 
+      case "response.completed":
+      case "response.incomplete":
+      case "response.failed": {
+        // Terminal lifecycle events carry the authoritative token accounting.
+        // An incomplete/failed response still consumed tokens, so its usage is
+        // recorded too rather than being dropped as a non-success.
+        usage = mapOpenAIResponsesUsage(event.response?.usage) ?? usage;
+        break;
+      }
+
       default:
-        // Ignore lifecycle (created/in_progress/completed), reasoning summaries,
-        // and non-function tool events.
+        // Ignore the remaining lifecycle events (created/in_progress),
+        // reasoning summaries, and non-function tool events.
         break;
     }
   }
+
+  return usage;
 }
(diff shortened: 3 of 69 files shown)


sroussey:
CI `test-vitest-unit` is red on `cd5f3a0f`, with **exactly one failure, inherited from the base branch**:

```
Test Files  1 failed | 389 passed | 1 skipped (391)
     Tests  1 failed | 4872 passed | 43 skipped (4916)
```

The failure is `packages/test/src/test/ai-provider/OpenAIGeminiCheckpointParams.test.ts` → "evicts and retries inline once on a NOT_FOUND", the same one reported on #641. Its stack runs entirely through `generateGeminiStreamWithCacheFallback`, and this PR's diff touches neither that function nor that test.

**Re-running will not help** — it is deterministic, not a flake. It clears only when the #641 question is decided:

- commit `5d9088f3` ("tighten reactive CachedContent NOT_FOUND matcher") made the matcher require a cache-scoped signal — a `cachedContents/…` `ResourceInfo` or `cachedContent` wording — specifically so a **model** 404 is not mistaken for a **cache** 404;
- the older test still throws `{ status: 404, code: "NOT_FOUND", message: "cache not found" }`, which carries neither signal, so the tightened matcher deliberately declines to evict and retry.

Either the fixture should become cache-scoped, or the matcher is stricter than the real Gemini error shape warrants. That second possibility is the reason this has not been "fixed" unilaterally: if a genuine cache eviction really does surface as an unscoped `NOT_FOUND`, loosening the fixture would paper over a production bug in which a stale `cachedContent` never triggers the inline-replay fallback.

Blocked on that decision. Everything else in this PR is green — build 84/84, build:types 41/41, and `provider-api` up from 413 to 431 tests.

---

PR #710, opened 2026-08-07
Migrate SQLite driver from better-sqlite3 to node:sqlite

sroussey:
Replaces the `better-sqlite3` dependency with the built-in `node:sqlite` module, stable in Node 24+ and shipped by Bun from 1.4.0. This drops a native dependency and collapses the Node and Bun SQLite drivers into one implementation.

## Key changes

- **Unified driver**: Node and Bun both run `node:sqlite` through a single `node.ts` entry point. The separate `bun.ts` driver and `@workglow/sqlite`'s `"bun"` export condition are gone; `BunExportConditions.test.ts` pins the two conditions that remain (both `@workglow/util`).
- **Error code translation**: maps `node:sqlite`'s generic `ERR_SQLITE_ERROR` + `errcode` onto better-sqlite3-style codes (`SQLITE_CONSTRAINT_UNIQUE`, `SQLITE_BUSY_RECOVERY`, …) so existing error handling keeps working — extended codes first, then the primary code from `errcode & 0xff`.
- **BigInt narrowing**: results are read as BigInt and narrowed back to `number` when the value round-trips exactly, avoiding `ERR_OUT_OF_RANGE` without losing precision on large keys.
- **Parameter binding**: `undefined` binds as SQL NULL, positionally and inside a named-parameter object, preserving the better-sqlite3 behaviour callers rely on for unset columns.
- **Transactions**: `node:sqlite` has no `transaction()`, so the driver implements BEGIN/COMMIT/ROLLBACK itself and downgrades a nested call to a SAVEPOINT. Nesting is decided by SQLite's own state, so a transaction opened through `exec("BEGIN")` or a prepared `BEGIN` is seen too.
- **Statement wrapper**: `NodeSqliteStatement` adapts `StatementSync` to the canonical `SqliteApi.Statement` contract.
- **Options**: `NodeSqliteOptions` keeps API compatibility and rejects renamed or unsupported options (`readonly`, `fileMustExist`, `verbose`, `nativeBinding`, `open`) rather than letting `DatabaseSync` ignore them — a leftover `readonly: true` would otherwise open the database read-write.

## `inTransaction` on the canonical API

`SqliteApi.Database` now declares a readonly `inTransaction`. Three nested-transaction guards — in `SqliteTabularStorage.runPutBulkOnHandle` and two in `SqliteAiVectorStorage` — used to probe for this member on the wrapper, where it was never declared, so each read `undefined` and the guards were dead code. A BEGIN opened by another storage sharing the connection was invisible, so a guard could issue a nested BEGIN; only the connection mutex kept that from surfacing.

Both drivers now answer from SQLite itself: the Node driver returns `DatabaseSync.isTransaction`, the browser (WASM) driver reads `sqlite3_get_autocommit` — the same primitive better-sqlite3 exposed under this name. Because that signal is trustworthy on both, the browser driver's `transaction()` also downgrades a nested call to a SAVEPOINT instead of issuing a second BEGIN, so the two implementations no longer diverge.

## Test coverage for the browser driver

The browser driver previously had none — no test imported it, and no vitest project configures a browser environment, so half of a two-implementation contract went unchecked. `@sqlite.org/sqlite-wasm` ships a Node entry, so the WASM driver runs under plain Vitest with no browser and no new harness. `SqliteBrowserDriver.contract.test.ts` covers `inTransaction`, savepoint nesting, async-body rejection, and commit/rollback. It loads the shipped browser bundle by path, since the package's `"./storage"` export resolves to the node driver wherever Vitest runs, and therefore needs the package built (as the rest of `packages/test` already does).

## Implementation details

- Error translation happens on throw via `translateError()` / `rethrow()`.
- The constructor closes the freshly opened handle if a post-open check fails, since a constructor that throws never hands the caller a `close()`.
- `close()` is idempotent, as it was on better-sqlite3 and `bun:sqlite`; `node:sqlite` throws `ERR_INVALID_STATE` on a second close. The latch is set only once the close has succeeded, so a transient failure still leaves a retry able to release the connection.
- Rows are rebuilt by spread, so a column named `__proto__` survives as an own property instead of hitting `Object.prototype`'s setter.
- `loadExtension()` rejects an entry point, which `node:sqlite` does not support.
- Version guards on `DatabaseSync.isTransaction` and `StatementSync.setReadBigInts` fail with a diagnosable message on an older `node:sqlite`.
- The async-transaction-body rejection is shared by both drivers, so a promise-returning body cannot silently commit at its first `await` on one platform while throwing on the other.

## Requirements

- `engines.node` is `>=24` on the root and on `@workglow/sqlite` — the dropped optional `better-sqlite3` peer was previously the only consumer-facing signal of a native floor.
- `engines.bun` is `>=1.4.0`, `packageManager` is `bun@1.4.0`, and every workflow pins `bun-version: 1.4.0`. Bun 1.4.0 is a tagged release, so the earlier canary pin is no longer needed; an exact pin also keeps CI reproducible.

## Verification

Run on Node 24.20.0 and Bun 1.4.0: build, lint, `typecheck:budget`, `typecheck:tests` and the test-discovery guard all clean; `storage` + `queue` is 2538 passing under vitest and green under the Bun runner; the full vitest unit and integration suites pass. The driver contract suites cover transactions and savepoint nesting, `inTransaction`, async-body rejection, option rejection, row prototypes, the `__proto__` column, BigInt narrowing, positional and named `undefined` binding, BLOB binding, error-code translation, foreign keys, WAL and the busy timeout.

https://claude.ai/code/session_01Xt5u21MrwDihp8osYYNQfd


sroussey:
CI is red on `test-vitest-integration`, but the failure is **pre-existing on `main`, not from this PR**. Every other check is green: `build`, `typecheck-budget`, CodeQL, and the other five vitest jobs.

**Same failure on the base commit.** [Run 31139502619](https://github.com/workglow-dev/libs/actions/runs/31139502619) on `main` at `8eb4452` — this PR's exact merge base, with `better-sqlite3` still in place — failed in the same job with the same assertion, same line, and the same totals:

| | main @ `8eb4452` | this PR @ `81e46e8` |
| --- | --- | --- |
| Failing test | `SqliteJobQueue > Same-process optimizations > deferred submit wakes before the poll interval elapses` | identical |
| Assertion | `expected 'TIMEOUT' not to be 'TIMEOUT'` @ `genericJobQueueTests.ts:766:26` | identical |
| Totals | 1 failed \| 1614 passed \| 42 skipped | identical |

**It also can't be reached by this diff.** The test times out waiting for the deferred-wake path, which is `JobQueueWorker.getIdleDelay()` reading `visible_at` — a **TEXT** column. The two driver behaviors this PR changes are BigInt narrowing (INTEGER columns only) and `undefined` → `null` on binding; neither touches a text timestamp read. The earlier assertions in the same test, which verify the stored `visible_at` value, pass — so the row is written correctly and only the 5s wake budget is missed.

Locally the test passes 3/3 standalone and again under full CPU saturation (all 4 cores pegged), so I could not reproduce it. On the CI runner the job reports 129s of test time inside an 83s wall clock, i.e. heavy parallel contention against a 5s budget.

Not fixing it here — it's an unrelated flaky timing assertion and a fix belongs in its own change. Happy to take it separately if you want it deflaked (raising the budget or gating the wake on an event rather than a deadline). I'll re-check this PR periodically and report when the job goes green.

For the record, the SQLite suites themselves are green on both runtimes: `storage` + `queue` under vitest/Node 24 is 90 files / 2376 tests passing, and under Bun canary 2375 passing with one unrelated pre-existing `JobErrorRegistry` logger-spy failure.

---

sroussey:
Closing the loop on my earlier note: **CI is fully green** — all 14 checks pass on `ba476f2` ([run #2904](https://github.com/workglow-dev/libs/actions/runs/31150486068)), including `test-vitest-integration`.

The branch is now rebased onto `main` @ `14e4130`, which picked up #712 / `3431d98` — the fix for the exact flaky test I flagged. That commit confirms the diagnosis independently: `getIdleDelay` saw an already-visible head job and returned the full poll interval, stranding claimable work. It surfaced through the Supabase fast-wake test too, so it was never SQLite-specific and nothing in this PR needed to change for it.

The rebase was conflict-free and the lockfile needed no changes. Local re-verification against the new base: `storage` + `queue` under vitest/Node 24 is 90 files / 2377 tests passing.

Leaving the PR in draft, since it was converted deliberately.

---

PR #719, opened 2026-08-07
Reorganize Tests and contracts

sroussey:
## Summary

Refactors test infrastructure across the monorepo to establish a consistent pattern for test-only exports. Test doubles (fakes, mocks, in-memory implementations) and conformance contracts now live in dedicated `test-entry.ts` files within each package, rather than in a centralized `@workglow/test` package. This improves modularity, reduces circular dependencies, and ensures test code can reference the same module instances as production code.

## Key Changes

- **New test entry points**: Added `src/test-entry.ts` to `@workglow/task-graph` and `@workglow/ai`, exporting test doubles and contracts from those packages
- **Streaming repository contract**: Extracted `streamingTaskOutputRepositoryContract.ts` as a reusable conformance suite for streaming task output repositories, eliminating ~95% duplication across three per-backend test suites
- **Pipe function task**: Moved `PipeFunctionTask` class from `Conversions.ts` to new `PipeFunctionTask.ts` to break a circular dependency (`Conversions` → `TaskRunner` → `Task` → `Conversions`)
- **Test discovery guard**: Added CI job (`test-discovery`) to catch unreachable test files that would silently skip due to section/kind filtering
- **Nightly Bun parity workflow**: Added scheduled workflow to surface Bun-specific issues without blocking PRs
- **Consolidated test doubles**: Moved `InMemoryTaskOutputRepository` from `@workglow/test` to `@workglow/task-graph/testing`, with similar consolidation for other test implementations
- **TypeScript test configs**: Added `tsconfig.test.json` to packages that define test files, enabling proper type checking of test code without affecting production builds
- **Updated imports**: Adjusted all test files to import test doubles from their new locations via package specifiers (e.g., `@workglow/task-graph/test`)

## Implementation Details

- Test doubles import their base classes by package specifier to ensure `instanceof` checks work correctly across module boundaries
- Conformance contracts use structural typing for repository interfaces rather than concrete types, allowing adapters to carry different additional properties
- The `test-entry.ts` pattern mirrors the public entry point structure, with test doubles defined locally and reset hooks/shared registries re-exported from an `_internal` bag on the public entry
- File reorganization preserves all test semantics; no test logic was changed, only relocated

https://claude.ai/code/session_01RT6KwvnW6VRuniHdpZiv7S


sroussey:
CI red on `typecheck-budget`, failing on the base branch too — not caused by this PR.

```
providers/supabase: 92297 instantiations vs budget 29527 (+213%)
```

Identical on `main`: run [#2921](https://github.com/workglow-dev/libs/actions/runs/31222080367) for `9135d06` ("feat(storage): enhance query operators…") fails with the same package, the same count (92297) and the same budget (29527). That commit changed `providers/supabase/src/storage/SupabaseTabularStorage.ts` and `packages/storage/src/tabular/ITabularStorage.ts` without re-baselining `scripts/typecheck-budget.json`.

This branch touches neither file. I merged `main` in at a1119af, which is how the failure arrived here.

I have deliberately **not** re-baselined it. A 3.2x jump in one package is large enough to be a genuine type-level regression from widening the operator union rather than expected growth, and burying that decision in a test-restructuring PR would both hide it and misattribute it. It wants either a look at `SupabaseTabularStorage`'s inference or an explicit `--update` from whoever owns the storage change.

Everything else on a1119af is green: `build`, `test-discovery`, `test-vitest-integration`, `test-vitest-rag`, all three provider jobs, CodeQL. Locally the merged tree runs 488 passed / 3 skipped (491 files), 5705 tests, exit 0.

Will re-run once the base recovers.

---

PR #963, opened 2026-09-20
Window list rows on the work in flight, not the head or tail

sroussey:
## Summary

Replaces the tail-pinning display strategy for truncated lists with a window that follows the work in flight. Instead of always showing the last N rows, the window now centers on running tasks with settled context above and pending work below, scrolling forward as tasks complete.

## Key Changes

- **New windowing algorithm** (`listWindow`, `windowStart`): Positions a fixed-size window on active rows, keeping one row of settled context above them and letting the window walk forward as tasks land. When nothing is running, anchors on the next pending task; when everything is done, shows the tail (where failures sort).

- **Row phase classification** (`rowPhase`, `RowPhase` type): Introduces three phases (settled/active/waiting) that work across both task and iteration slot vocabularies, enabling the window to identify where the work is regardless of status naming.

- **Dual summary lines**: Lists can now hide rows at both ends of the window. Added `HIDDEN_ABOVE_GLYPH` (▲) and `HIDDEN_BELOW_GLYPH` (▼) to distinguish "work behind us" from "work to come" in the summary text.

- **Updated viewport planning**: `listRows` now accounts for summary lines at both ends; `countHidden` measures what the window actually hides rather than what the cap would hide, so a list that keeps every row because the summary line isn't worth its space reports zero hidden.

- **React hook simplification**: `useVisibleRows` now returns a `ListWindow<T>` with `visible`, `before`, and `after` fields instead of `visible` and `hidden`, making the two-ended nature explicit at the call site.

- **Display updates**: `SubtaskRows` and `RootTaskRows` now render hidden-above and hidden-below lines with appropriate glyphs, replacing the single hidden line.

## Implementation Details

The window anchors on the running rows (the frontier of work in flight). A span of running tasks wider than the window keeps its head—those started first and are likeliest to finish next. With nothing running, it anchors on the next pending task (the moment before a graph starts or between steps). With nothing left to run, it shows the tail (the end of a finished list is the last thing it did, and failures sort there).

The window keeps one settled row of context above the frontier (`LEAD_CONTEXT_ROWS = 1`), which is enough to show the pipeline is moving rather than stuck. This context walks with the window as tasks land.

A list one row over its cap keeps that row rather than spending the space on a summary line, since the real row is the better use of the space.

Updated documentation in `runViewport.ts` explains the three rules shaping the viewport: the live region never shrinks, a list shows the work in flight, and depth pays for overflow.

https://claude.ai/code/session_01N1mYmvAD8KHcwh8sNyUqTc


sroussey:
This PR now carries a second commit (`0c0cd95`) outside its stated scope, in `packages/storage/` — flagging it so the storage change isn't a surprise in review.

**What was failing:** the `lint` job, on `packages/storage/src/tabular/SqlTabularMigrationApplier.ts:226` —

```
error typescript(no-base-to-string): 'item' may use Object's default
stringification format ('[object Object]') when stringified.
```

**It isn't this PR's.** The viewport change touches only `examples/cli/src/ui/**` and `.claude/CLAUDE.md`. The same `lint` job is red on the base commit `75ca8c6` for the same rule ([run 35493338274](https://github.com/workglow-dev/libs/actions/runs/35493338274) — `lint` failed, every other job passed), and `75ca8c6` is the commit that introduced `arrayLiteralSql`.

**Fixed here rather than reported**, at the author's request. `arrayLiteralSql` already validates that every element is `string | number | boolean | null` and throws otherwise, but the check only reached the control flow, so the render below still saw `unknown`. The rule is right on the merits: `String()` on an object would write `[object Object]` as a column default onto every existing row. The fix collects into a typed array behind a predicate, so the type says what the loop had already established — same validation set, same error text, same rendering.

Verified by reverting the fix and re-running the exact CI command: `bunx oxlint --type-aware --report-unused-disable-directives --deny-warnings` exits 1 with that error without it, 0 with it (against real `dist/*.d.ts` from `bun run build:types`). `bun run format-check` clean; storage-tabular, tabular-migrations, storage-migrations and the CLI UI suites pass (620 tests).

---

PR #988, opened 2026-09-21
test(storage): hold ScopedTabularStorage to the tabular and join contracts

### LINKED ISSUE #975: `FsFolderTabularStorage` and `ScopedTabularStorage` run no join assertion of any kind — two of the backends #947 named, and one of them has a bespoke `join` implementation
## What

`a0dbefbbc` wired `runTabularJoinContract` (16 semantics cases + the strategy-parity block + bounded-left-read) into ten test files, which is the substance of #947. Two backends the issue named were not among them.

```
$ for f in packages/test/src/test/storage-tabular/*.test.ts; do
    printf '%-55s contract=%s join=%s generic=%s\n' "$(basename $f)" \
      "$(grep -c runTabularStorageContract $f)" \
      "$(grep -c runTabularJoinContract $f)" \
      "$(grep -c runGenericTabularStorageTests $f)"; done
…
FsFolderTabularStorage.integration.test.ts             contract=2 join=0 generic=2
ScopedTabularStoragePostgres.integration.test.ts       contract=0 join=0 generic=0
ScopedTabularStorageSqlite.integration.test.ts         contract=0 join=0 generic=0
…
```

Every other tabular backend has `join=2` or more.

## Why each one matters

**`FsFolderTabularStorage`** inherits `BaseTabularStorage.join` — the hash fallback — and was one of the seven backends #947 listed as having no join assertion. It rides `runTabularStorageContract` and `runGenericTabularStorageTests`, so adding the join contract is one call. It is also the backend whose left read is genuinely expensive: `join bounded left read` asserts that a bounded join stops reading the left side early, and on a folder-per-row storage "reads the whole left side" is one `readFile` per row.

**`ScopedTabularStorage`** (`packages/knowledge-base/src/knowledge-base/ScopedTabularStorage.ts`) is the stronger case, because it does **not** inherit — it has its own delegate:

```
packages/knowledge-base/src/knowledge-base/ScopedTabularStorage.ts:190
   * side's own on the right, both applied in the join condition so a LEFT
   * JOIN keeps its unmatched rows. The inner storages are handed to the join
```

Splicing the scope predicate into the join condition rather than applying it after is exactly the rule `a right-side filter under a left join does not drop unmatched left rows` states, and there is no shared case anywhere asserting the scoped wrapper gets it right. `ScopedVectorStorage` already rides the shared vector contract (`ScopedVectorStorage.contract.test.ts`); its tabular sibling rides neither the tabular contract nor the join one.

## Proposed fix

1. Add `runTabularJoinContract(…)` to `FsFolderTabularStorage.integration.test.ts` beside the two suites it already calls.
2. Give `ScopedTabularStorage` a contract caller — `runTabularStorageContract` + `runTabularJoinContract` over a scoped pair, the way `ScopedVectorStorage.contract.test.ts` does for vectors. A wrapper that rewrites a predicate is the implementation a shared contract exists for.

Measured on `origin/main` @ `d98940f33` (0.6.7 + 4). Residual of #947. Found during the 2026-09-21 review of `packages/`.

sroussey:
Closes #975.

## `ScopedTabularStorage` — the real half, and it found two bugs

The issue is right that this is the stronger case: the wrapper does **not** inherit `BaseTabularStorage.join`. It has its own delegate that splices the scope predicate into the join **condition** rather than applying it afterwards — which is exactly what the contract's `a right-side filter under a left join does not drop unmatched left rows` states — and nothing asserted it. `ScopedVectorStorage` already rides the shared vector contract; its tabular sibling rode neither the tabular one nor the join one.

New `packages/test/src/test/storage-tabular/ScopedTabularStorage.contract.test.ts` runs both contracts over a scoped pair (`InMemoryTabularStorage` inners keyed `["kb_id", …]`), following the `ScopedVectorStorage.contract.test.ts` pattern. `expectSqlPushdown` is left unset per the option's own docstring — a wrapper hands its **inner** storage to the join, so the right-side `query` lands on the inner's.

### Bug 1 — the contract found it on its first run

`deleteSearch` injected `kb_id` before delegating:

```ts
await this.inner.deleteSearch({ ...criteria, kb_id: this.kbId });
```

To the inner storage, `kb_id` is just another column that matches some rows and not others. So `shouldRunDeleteSearch` saw a narrowed predicate and let the delete run — and `{ col: { value: [], operator: "not-in" } }` (an exclusion list built from caller input that came back empty) emptied **the whole scope**, which is the entire universe this wrapper's caller can see. That is precisely the outcome the unfiltered-delete guard exists to refuse, one level down, and the wrapper's own bookkeeping was what defeated it.

Fixed by screening the **caller's** criteria before injecting. Reverting just the wrapper reproduces the failure:

```
 × deleteSearch refuses criteria that reduce to an empty exclusion
      Tests  1 failed | 39 passed | 134 skipped
```

### Bug 2 — from the Copilot review, verified

*"The bulk-delete path does not emit the required `delete` event."* Correct. Every concrete implementation announces a bulk `deleteSearch` — `InMemoryTabularStorage` emits one per matched row, and Postgres, SQLite, DuckDB, Supabase and IndexedDB each emit one built by `BaseTabularStorage.deleteIdentity`. The scoped wrapper emitted nothing.

It owns an emitter separate from the inner's, so the inner's event reaches the inner's listeners and never the wrapper's: a scoped caller saw keyed `delete` and `clearall`, and silence for the path a scoped caller actually deletes through — which `InMemoryTabularStorage.delete`'s own comment states:

> A keyed delete emits only the key. Bulk `deleteSearch` carries the matched row … and **scoped callers delete via `deleteSearch`** so the owner columns reach subscribers without a read-back.

The identity is built from the caller's criteria, before `kb_id` is injected — that column is the wrapper's bookkeeping, and everything else the wrapper hands back has it stripped.

## Shared helpers, so a wrapper can meet the contract

A wrapper implementing `ITabularStorage` without extending `BaseTabularStorage` can reach neither a module-private function nor a protected member, so both fixes needed the rule to be reachable:

- `shouldRunDeleteSearch` is exported from `@workglow/storage`
- `deleteIdentity`'s body moves to `deleteSearchIdentity` in `tabularValidation.ts` and is exported beside it; the base method delegates, so all five backends are unchanged

## `FsFolderTabularStorage` — not needed, and the premise is wrong

The issue says FsFolder "inherits `BaseTabularStorage.join` — the hash fallback", so adding `runTabularJoinContract` is one call. It does not inherit it:

```
packages/storage/src/tabular/FsFolderTabularStorage.ts:429
  override async join<R, T extends JoinType>(...): Promise<JoinedRow<Entity, R, T>[]> {
    throw new StorageUnsupportedError("join", "FsFolderTabularStorage");
```

`join` is overridden to refuse, for the same reason `query` is (`:419`) — and the hash fallback needs `query` on the right side, which this backend does not have. Its integration test already pins the refusal, so `join=0` in the issue's grep correctly reflects a backend with no join to conform to, and the `join bounded left read` case can't apply since the left read never happens. Left alone rather than adding a suite that would be skipped wholesale.

## Tests

Beyond the two shared contracts, five cases for what this class adds rather than what the interface promises: both sides scoped with `kb_id` stripped; unmatched left rows kept under a LEFT join; an unscoped right side refused; the bulk delete emitting exactly one `delete` with no `kb_id` on the identity; and neither guard-stopped path emitting.

```
storage-tabular + storage + knowledge-base:  43 passed | 2 skipped · 1086 tests
storage (integration):                       26 passed · 1968 tests
build:types 47/47 · oxlint --type-aware clean · typecheck-budget OK · check-sections OK
```

## Follow-up worth its own issue

The shared tabular contract does not assert the bulk-delete `delete` event for **any** backend — which is why this suite went green while the wrapper dropped it, and why the gap survived five conforming backends. Adding the assertion centrally would close it for every adapter at once, but it changes expectations across all of them, so it is not in this PR.

https://claude.ai/code/session_01Dxc97qR2axPNmoQPVJ3yJm


sroussey:
The Copilot review's one note — *"the bulk-delete path does not emit the required `delete` event"* — is correct. Verified and fixed in 92b3556.

**What I checked.** Every concrete implementation announces a bulk `deleteSearch`:

- `InMemoryTabularStorage` emits one `delete` per matched row, with the full entity
- `PostgresTabularStorage`, `SqliteTabularStorage`, `DuckDbTabularStorage`, `SupabaseTabularStorage` and `IndexedDbTabularStorage` each emit one built by `BaseTabularStorage.deleteIdentity`

`ScopedTabularStorage` emitted nothing. It owns an emitter separate from the inner storage's, so the inner's event reaches the inner's listeners and never the wrapper's — a caller subscribed to a scoped storage saw keyed `delete` and `clearall`, and nothing for `deleteSearch`. That is the path a scoped caller actually deletes through, which `InMemoryTabularStorage.delete`'s own comment states outright:

> A keyed delete emits only the key. Bulk `deleteSearch` carries the matched row … and **scoped callers delete via `deleteSearch`** so the owner columns reach subscribers without a read-back.

Pre-existing rather than introduced here, but it is in the method this PR already changed, and holding this wrapper to the contract is what the PR is for.

**The fix.** `deleteSearch` now emits `delete` after the inner call, with the identity built from the **caller's** criteria — before `kb_id` is injected. That column is the wrapper's bookkeeping rather than one the caller named, and everything else the wrapper hands back has it stripped, so leaking it here would be inconsistent.

`deleteIdentity`'s body moves to `deleteSearchIdentity` in `tabularValidation.ts`, exported beside `shouldRunDeleteSearch`, because a wrapper implementing `ITabularStorage` without extending `BaseTabularStorage` cannot reach a protected member. The base method delegates to it, so all five backends are byte-identical in behaviour.

**Tests.** Two cases: the bulk path emits exactly one `delete` whose identity is `{ tenant: "t1" }` with no `kb_id`; and neither guard-stopped path emits — empty criteria are a no-op, and an all-empty exclusion throws, which is not a delete. Removing the emit line fails the first and nothing else.

```
storage-tabular + storage + knowledge-base:  43 passed | 2 skipped · 1086 tests
storage (integration):                       26 passed · 1968 tests
build:types 47/47 · oxlint --type-aware clean · typecheck-budget OK
```

Worth noting separately: the shared tabular contract does not assert this event for any backend, which is why the suite went green without it. Adding it centrally would be a good follow-up, but it touches every adapter's expectations so I've kept it out of this PR.

---

